PR #1, opened 2022-10-20
Develop

tiserett:
Finished server
Created services folder for files with helpers functions
Implemented routes


Anton-Liada (approved):
🚀🚀

src/routers/expenseRouter.js
```diff
@@ -0,0 +1,150 @@
...
+
+  const foundExpense = expenseServise.findById(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  expenseServise.update(foundExpense, req.body);
```
AlinaMatuschak, line 128: can you rewrite the update: send just the expense id, and request body?

src/routers/expenseRouter.js
```diff
@@ -0,0 +1,150 @@
...
+
+  const foundExpense = expenseServise.findById(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  expenseServise.remove(expenseId);
```
AlinaMatuschak, line 146: you can throw an error inside the remove method and catch it

```suggestion
  try {
    expenseServise.remove(expenseId);
  } catch (e) {
    res.sendStatus(404);

    return;
  }
```

AlinaMatuschak (changes requested):
Awesome 👏 . But should create controllers for expenses and users )

src/routers/expenseRouter.js
tiserett: If I'm not mistaken Misha said that these methods should be clean, and we should check everything inside our requests, and for me it's easier to keep it this way

src/routers/expenseRouter.js
tiserett: what's the problem with this implementation? I find my expense, then I call update method to update it with certain data

AlinaMatuschak: - you find expense just for updating it (so it needs just for the update): when you will add db, you use just id, so it will more easier for you
- if you will use the update method several times, you need each time to add duplicated code to find expense that you need to update

src/routers/expenseRouter.js
AlinaMatuschak: Okay, if you are more comfortable using your way, you can check existing expense before calling remove method )

src/contolers/expenses.js
```diff
@@ -0,0 +1,154 @@
+'use strict';
+
+const expenseServise = require('../services/expenses.js');
+const userServise = require('../services/users.js');
+
+let nextId = 0;
```
AlinaMatuschak, line 6: better move it to services or create a helper that generate random number

src/contolers/users.js
```diff
@@ -0,0 +1,108 @@
...
+function getAll(req, res) {
+  const users = userServise.getAll();
+
+  res.statusCode = 200;
+
+  if (!users) {
+    res.send([]);
+
+    return;
+  }
```
AlinaMatuschak, line 36: is getAll return always array?
```suggestion
  const users = userServise.getAll();

  res.statusCode = 200;
```

AlinaMatuschak (approved):
👍

PR #5, opened 2022-10-20
Sokution

kir8mir:


src/createServer.js
```diff
@@ -1,13 +1,230 @@
...
+      res.setHeader('Content-Type', /application\/json/);
+      res.sendStatus(404);
+      res.end();
+    }
+  });
+
+  app.post('/users', (req, res) => {
+    const { name } = req.body;
+    const newUser = {
+      id: 1,
```
Andrii-Stiebalov, line 45: O my, it's _**'kostyl'**_ . Use thing like  `id: newExpensesId++`  with closure😉

src/createServer.js
```diff
@@ -1,13 +1,230 @@
...
+      name,
+    };
+
+    if (!name) {
+      res.statusCode = 400;
+      res.end();
+      users = [];
+    } else {
```
Andrii-Stiebalov, line 53: you don't need `else` if you `return` the value in the first `if`

src/createServer.js
```diff
@@ -1,13 +1,230 @@
...
+
+    res.statusCode = 404;
+
+    users.map((user, index) => {
+      if (user.id === id) {
+        users.splice(index, 1);
+        res.statusCode = 204;
+      }
+    });
+
```
Andrii-Stiebalov, line 91: `map` is not the best method in this case, use `forEach` and `break` the loop after the `if` is done

src/createServer.js
```diff
@@ -1,13 +1,230 @@
...
+    res.statusCode = 200;
+    res.send(expenses);
+  });
+
+  app.get('/expenses/:expenseId', (req, res) => {
+    const id = parseInt(req.path.slice(1).split('/')[1]);
+
+    const expenseItem = expenses.find(expense => expense.id === id);
+
+    console.log(212, expenseItem);
```
Andrii-Stiebalov, line 146: delete `console` 👨‍💻

Andrii-Stiebalov (changes requested):
Good job bro. But fix those comments. And better solve this problem with express.Router() and expres.cors(), because it is likely that the work will be rejected without this✨👨‍💻🚀

Andrii-Stiebalov (approved):
Good job!!!🚀

src/controllers/expenses.js
```diff
@@ -0,0 +1,154 @@
...
+
+  if (!title || !userServise.exist(userId)) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newExpense = {
+    ...req.body,
+    id: nextId++,
```
dellachaise, line 22: Check is id this way works properly? Maybe better use `Math.random with new Date` and count it here. No initializing in line 6(because as I understand, it doesn't work like with closure). (We could discuss it 🙂 )

src/controllers/expenses.js
```diff
@@ -0,0 +1,154 @@
...
+
+  const foundExpense = expenseServise.findById(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.statusCode = 200;
```
dellachaise, line 104: This line is redundant, delete it. And delete it all over the project.

200 status code - is the default one. 

[Read more details here](https://stackoverflow.com/questions/53403078/express-is-it-necessary-to-respond-with-status-200)

src/controllers/expenses.js
```diff
@@ -0,0 +1,154 @@
...
+  const foundExpense = expenseServise.findById(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  expenseServise.remove(expenseId);
+  res.sendStatus(204);
```
dellachaise, line 145: Catch an error. What if something went wrong with removing?
 
src/controllers/users.js
```diff
@@ -0,0 +1,108 @@
+'use strict';
+
+const userServise = require('../services/users.js');
+
+let nextId = 0;
```
dellachaise, line 5: The same as in the previous file.

src/controllers/users.js
```diff
@@ -0,0 +1,108 @@
...
+    return;
+  }
+
+  res.statusCode = 200;
+
+  res.send(foundUser);
+};
+
+function update(req, res) {
+  const userId = Number(req.params.id);
```
dellachaise, line 64: Try this way 👇 
```suggestion
  const { userId: Number(userId)  } = req.params;
```

Using destructure - is a better option here. And in future test tasks when it possible.

src/createServer.js
```diff
@@ -1,11 +1,24 @@
+/* eslint-disable no-console */
```
dellachaise, line 1: Delete it.

Using `eslint-disable` is a bag practice. Try to avoid it here and in test tasks. Find a way to solve it without `eslint-disable`.(here I understand that it is redundant because you already deleted console.log).

But I need to worn you - how you should check you code in future test tasks.

dellachaise (approved):
Good job 👍 

Need to add some improvements(read comments).

PR #9, opened 2022-10-21
Solution

Jamal-Amoian:


Jamal-Amoian:
> _ No description provided. _

I don't really know what is that problem, I started these tests separately from each other and it works, I think, that the problem in the test

andrii-dmytruk (approved):
Well done

Jamal-Amoian:
https://imgur.com/a/rArOXM5

src/routes/usersRoute.js
```diff
@@ -0,0 +1,80 @@
...
+const cors = require('cors');
+
+const router = express.Router();
+
+router.use(cors());
+
+router.get('/', (req, res) => {
+  res.statusCode = 200;
+  res.send(services.getUsers());
+});
```
ArturErmolenko, line 14: it will be great to transfer logic from router, in controllers you could create function
```
function getAll(req, res) {
   const users = usersServise.getAll();

   res.statusCode = 200;
   res.send(users);
 }
```
and your route will look like this
`router.get('/', usersControllers.getAll);`

you also need to change other routes, then the architecture of your application will be correct

ArturErmolenko (changes requested):
Great job 🚀 

let's try to connect controllers file for users and expenses

ArturErmolenko (approved):
Great 🚀 Well done 👍

PR #10, opened 2022-10-21
accounting app Node.JS

yungdena:


lizapolokhina (approved):
looks good)

src/createServer.js
```diff
@@ -1,11 +1,164 @@
...
+
+    users.push(newUser);
+    res.statusCode = 201;
+    res.send(newUser);
+  });
+
+  app.get('/users', (req, res) => {
+    if (!users.length) {
+      res.statusCode = 200;
+      res.send([]);
```
dellachaise, line 34: Why did you send `[]`, could you simply `sendStatus` here?

src/createServer.js
```diff
@@ -1,11 +1,164 @@
...
+    }
+
+    const newUser = {
+      id: Math.floor(Math.random() * 1000),
+      name,
+    };
+
+    users.push(newUser);
+    res.statusCode = 201;
+    res.send(newUser);
```
dellachaise, line 28: here use use `res.statusCode + res.send` in code below `(line 49-50) `res.sendStatus + res.send` in kind of the same case(when you use `send` not only status send). Better do with the same logic all over the code. Firs option - `res.statusCode + res.send` as for me - preferable.

src/createServer.js
```diff
@@ -1,11 +1,164 @@
...
+  };
+
+  app.delete('/users/:userId', (req, res) => {
+    const filteredUsers = filterById(users, req.params.userId);
+
+    if (users.length === filteredUsers.length) {
+      res.sendStatus(404);
+    }
+    users = filteredUsers;
+    res.sendStatus(204);
```
dellachaise, line 77: Catch an error. What if something went wrong with removing?
 
src/createServer.js
```diff
@@ -1,11 +1,164 @@
...
+  });
+
+  app.post('/expenses', (req, res) => {
+    const { userId, spentAt, title, amount, category, note } = req.body;
```
dellachaise, line 81: You need here only `userId` 
That's why better rewrite 👇 
```suggestion
    const { userId } = req.body;
```
and in `newExpense` you could just pass `...req.body`.

tests/expense.test.js
```diff
@@ -188,6 +188,7 @@ describe('Expense', () => {
         });
 
       const response = await api
+      // eslint-disable-next-line
```

dellachaise (changes requested):
Good start 👍
Just create routes, controllers and services (split your code according to [MVC](https://mate.academy/learn/nodejs/express-nodejs#/video/2723))

PR #12, opened 2022-10-21
basic version

Alzay007:


src/createServer.js
```diff
@@ -1,11 +1,242 @@
...
+    const { name } = req.body;
+
+    if (!name) {
+      res.sendStatus(400);
+
+      return;
+    }
+
+    const newUser = {
+      id: Math.max(0, ...users.map(({ id }) => id + 1)),
```
Desertiks, line 50: You can just create variable with number 1 and then just increase the value by 1
```
let idCounter = 1;
const newUser = {
id : idCounter++,
}
```

src/createServer.js
```diff
@@ -1,11 +1,242 @@
...
+
+  app.post('/expenses', express.json(), (req, res) => {
+    const {
+      userId,
+      spentAt,
+      title,
+      amount,
+      category,
+      note,
+    } = req.body;
```
Desertiks, line 172: You don't need to destructing it like that, you can just write:
```
const newExpense = {
...req.body,
}
```

src/createServer.js
```diff
@@ -1,11 +1,242 @@
...
+
+  let users = [];
+  let expenses = [];
+
+  app.use(cors());
+
+  app.get('/users', (req, res) => {
+    if (!users) {
+      res.statusCode = 200;
+
```
Desertiks, line 17: I guess it's unnecessary to do this check, if your users is empty we can just turn our array back.

Desertiks (changes requested):
Fix comments below and split createServer on services, controllers and routes but I think you know that you need to do that without this comment :)

src/controllers/expenses.js
```diff
@@ -0,0 +1,143 @@
...
+  filterById,
+  getExpenseById,
+  createExpense,
+  removeExpense,
+  updateExpense,
+} = require('../services/expenses.js');
+const userService = require('../services/users.js');
+const { allUsers } = require('../controllers/users.js');
+
+let expenses = [];
```
Desertiks, line 16: Expenses array and method that clean expenses `clearExpensesArray` must be in services
This also applies to users

Desertiks (changes requested):
It looks great 🔥 
Just move expenses and users arrays

src/controllers/expenses.js
```diff
@@ -0,0 +1,138 @@
...
+  const expenses = takeExpenses();
+  const newExpenses = removeExpense(expenseId);
+
+  if (expenses.length === newExpenses.length) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  // removeExpense(expenseId);
```
Desertiks, line 104:
```suggestion

```

Desertiks (approved):
Nice! 👍

src/controllers/expenses.js
dellachaise, line 104: I think you need to uncomment this line.  What do you think?

src/services/expenses.js
```diff
@@ -0,0 +1,85 @@
+'use strict';
+
+let nextExpenseId = 1;
```
dellachaise, line 3: I think, more reliable in this case would  use Math.random multiply with new Date inside of the function where you use it.

dellachaise (approved):
Good job 👍 

See comments for adding some improwmwnts.

PR #13, opened 2022-10-21
Accounting app v2 implemented

andrii-dmytruk:


Jamal-Amoian (approved):
Good job!

src/controllers/users.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
+
+const userService = require('../services/users.js');
+
+function getAll(req, res) {
+  const users = userService.getAll();
+
```
ArturErmolenko, line 7: you should add status 200

src/controllers/users.js
```diff
@@ -0,0 +1,83 @@
...
+function getOne(req, res) {
+  const { userId } = req.params;
+  const foundUser = userService.getById(+userId);
+
+  if (!foundUser) {
+    res.send(404);
+
+    return;
+  }
+
```
ArturErmolenko, line 20: you should add conditions if userId didn't pass for status 400

ArturErmolenko (changes requested):
Great job 🚀 
Just fix comments with statuses, check all cases and return 400 with any message if required parameter is not passed.

ArturErmolenko (approved):
Well done 🔥

PR #15, opened 2022-10-21
V1 Solution

RomanYakm:


src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+  app.get('/users/:userId', (req, res) => {
+    const { userId } = req.params;
+    const foundUserId = usersData.find(user => user.id === +userId);
+
+    if (!foundUserId) {
+      res.status(404);
+    }
+
+    res.send(foundUserId);
+  });
```
ibondrk, line 61: Also u should check is your request correct

```suggestion
  app.get('/users/:userId', (req, res) => {
    const { userId } = req.params;
    
    if (isNaN(+userId)) {
      res.sendStatus(400);

      return;
    }
    
    const foundUserId = usersData.find(user => user.id === +userId);

    if (!foundUserId) {
      res.status(404);
    }

    res.send(foundUserId);
  });
```

image

src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+
+  app.post('/users', express.json(), (req, res) => {
+    const { name } = req.body;
+
+    const newUser = {
+      id: Math.random(),
+      name,
+    };
```
ibondrk, line 69: It would be nice if you assigned your `id`s in some more predictable order. 
ex: 
create a variable at the begin:
```
  let nextUserId = 1;
```

and just increase it where you need:

```
    const newUser = {
      id: nextUserId++,
      name,
    };
```

src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+
+    if (!name) {
+      res.status(400);
+      res.end();
+    } else {
+      usersData.push(newUser);
+      res.setHeader('Content-Type', 'application/json');
```
ibondrk, line 76: 🔥

src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
 
 function createServer() {
   // Use express to create a server
   // Add a routes to the server
   // Return the server (express app)
+  const app = express();
+
+  let usersData = [];
+  let exprensesData = [];
+
```
ibondrk, line 13: You can add 
`app.use(express.json());`
instead of using `express.json()` in every function

src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+      res.status(201);
+      res.send(newExprenses);
+    }
+  });
+
+  app.get('/expenses', (req, res) => {
+    const queryCheck = Object.keys(req.query).length;
+
+    if (!exprensesData) {
+      res.send([]);
```
ibondrk, line 107: if to be honest i don't understand what exectly u check. is Array is empty? if it's any way your funtion will return it

src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+    } else if (queryCheck === 0) {
+      res.setHeader('Content-Type', 'application/json');
+      res.status(200);
+      res.send(exprensesData);
```
ibondrk, line 111: this part you can send just once in the end of function after all checks

src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+    } else {
+      if (req.query.userId) {
+        const filterDataUserId = exprensesData.filter(
+          exprens => exprens.userId === +req.query.userId);
+
+        if (req.query.category) {
+          const filterDataCategory = exprensesData.filter(
+            exprens => exprens.category === req.query.category);
+
+          res.send(filterDataCategory);
```
ibondrk, line 121: will be good to use destructure below the same

src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+      };
+
+      exprensesData.push(newExprenses);
+      res.setHeader('Content-Type', 'application/json');
+      res.status(201);
+      res.send(newExprenses);
+    }
+  });
+
+  app.get('/expenses', (req, res) => {
```
ibondrk, line 103: generally try to restructure your code so that there are fewer complex conditions. It's hard to read now

ibondrk (changes requested):
Good one, just need to add some fixes) Sure u can make it better)
And one more general think. Would be good to split this ''one big function''

PR #21, opened 2022-10-21
accounting

NataliyaAn:
Не завершено!!! Доробляю


src/expensePoints.js
```diff
@@ -0,0 +1,167 @@
...
+    const deleteExpense = expenses.find(exp => exp.id === +id);
+
+    if (!deleteExpense) {
+      res.sendStatus(404);
+      res.send('Expense is not found');
+
+      return;
+    }
+
+    expenses.splice(expenses.indexOf(deleteExpense), 1);
```
max-kravchenko, line 160: Great idea with splice!

src/userPoints.js
```diff
@@ -0,0 +1,103 @@
...
+    users.push(newUser);
+
+    res
+      .status(201)
+      .send(newUser);
+  });
+
+  app.delete('/:userID', (req, res) => {
+    const { userID } = req.params;
+    const deleteUser = users.find(user => user.id === +userID);
```
max-kravchenko, line 86: your variable sounds like action. deletedUser would sound better as a variable.

src/expensePoints.js
```diff
@@ -0,0 +1,167 @@
...
+    expenses.push(newExpense);
+
+    res
+      .status(201)
+      .send(newExpense);
+  });
+
+  app.delete('/:id', (req, res) => {
+    const { id } = req.params;
+    const deleteExpense = expenses.find(exp => exp.id === +id);
```
max-kravchenko, line 151: this variable sounds like action as well. deletedExpense would be better for understanding)

src/expensePoints.js
```diff
@@ -0,0 +1,167 @@
+'use strict';
+
+function expensePoints(app, { users }, { expenses }) {
+  let newExpenseID = 1;
+
+  app.get('/', (req, res) => {
+    const { userId, category, from, to } = req.query;
+
+    let getExpenses = [...expenses];
```
max-kravchenko, line 9: getExpenses sounds like function. But it is a variable. MB it is better to call it receivedExpenses.. Smth like that

max-kravchenko (approved):
Great job! Only minor comments below.

p.s. to pass tests you should fix a length of line in `/home/runner/work/node_accounting-app-v2/node_accounting-app-v2/tests/expense.test.js
Error:   191:1  error  Line 191 exceeds the maximum line length of 80  max-len`

p.p.s. I am not sure but mentors could ask you to split the app into more files (like controllers and services from Misha's videos)

src/expensePoints.js
```diff
@@ -0,0 +1,167 @@
...
+  });
+
+  app.get('/:id', (req, res) => {
+    const { id } = req.params;
+    const foundExpense = expenses.find(expense => expense.id === +id);
+
+    if (!id) {
+      res
+        .status(400)
+        .send('ID of expense is requered');
```
ArturErmolenko, line 46: Do not rely on the unknown string, make constants for this.

ArturErmolenko (changes requested):
Great job 🔥 

```
function getAll() {
   return users;
 }
```
and function on you controllers file

```
function getAll(req, res) {
   res.status(200)
   res.send(users);
 }
```
than your route will look like this

Also I agree with comments about variables naming, change it where it needed

PR #27, opened 2022-10-23
Solution - All in one file

OlgaOchichenko:


KorniychukLesya (approved):
Good job!

src/createServer.js
```diff
@@ -1,11 +1,246 @@
...
+    res.statusCode = 201;
+    res.send(newUser);
+  });
+
+  app.get('/users', express.json(), (req, res) => {
+    if (!users.length) {
+      res.send([]);
+
+      return;
+    }
```
dellachaise, line 34: Why did you do this check?( I asked, because just interested about purpose, not saying that it is wrong, just trying to understand)

src/createServer.js
```diff
@@ -1,11 +1,246 @@
...
+      res.sendStatus(404);
+
+      return;
+    }
+
+    Object.assign(foundUser, { name });
+
+    res.statusCode = 200;
+
+    res.send(foundUser);
```
dellachaise, line 84: Catch an error. What if something went wrong with updating?

src/createServer.js
```diff
@@ -1,11 +1,246 @@
...
+
+    if (filterUser.length === users.length) {
+      res.sendStatus(404);
+
+      return;
+    }
+
+    users = filterUser;
+
+    res.sendStatus(204);
```
dellachaise, line 99: Here the same.

src/createServer.js
```diff
@@ -1,11 +1,246 @@
...
+
+  app.get('/expenses', express.json(), (req, res) => {
+    const {
+      userId,
+      category,
+      from,
+      to,
+    } = req.query;
+
+    const idNumber = +userId;
```
dellachaise, line 129: No need to create a separate variable for this purpose, because it is a too simple operation. You could just use `+` or `Number` when it needed.

src/createServer.js
```diff
@@ -1,11 +1,246 @@
...
+    }
+
+    res.statusCode = 200;
+    res.send(expenses);
+  });
+
+  app.get('/expenses/:id', express.json(), (req, res) => {
+    const { id } = req.params;
+
+    const idNumber = Number(id);
```
dellachaise, line 183: Here the same. And better use the same way converting to number all over the project  - `+` or `Number`.

dellachaise (changes requested):
Great start 👍
Just create routes, controllers and services (split your code according to [MVC](https://mate.academy/learn/nodejs/express-nodejs#/video/2723))

PR #28, opened 2022-10-23
accounting app was done

vladgusakovskiy:


Anastasiia-Khmelovska (approved):
great, but maybe you need to divide it to separate files

vladgusakovskiy:
yes, I'm going to do this

ArturErmolenko (changes requested):
Well done 🔥 
Lets try to separate logic in different files (services & controllers & route)

PR #29, opened 2022-10-23
server for accounting app was created

Andrii-Stiebalov:
👋🌝


kir8mir (approved):
Seems perfect, GJ!

src/controllers/expenses.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
+
+const { expensesServices } = { ...require('../services/expenses.js') };
+const { userServices } = { ...require('../services/users') };
```
AlinaMatuschak, line 4:
```suggestion
const { expensesServices } = require('../services/expenses.js');
const { userServices } = require('../services/users');
```

src/controllers/expenses.js
```diff
@@ -0,0 +1,101 @@
...
+
+const expenseController = {
```
AlinaMatuschak, line 6: better use classes if you don’t want to use functions

src/controllers/expenses.js
```diff
@@ -0,0 +1,101 @@
...
+  getAll: (req, res) => {
+    const userId = req.query.userId;
+    const category = req.query.category;
+    const from = req.query.from;
+    const to = req.query.to;
```
AlinaMatuschak, line 11: Don’t forget about destructuring

src/controllers/expenses.js
```diff
@@ -0,0 +1,101 @@
...
+      vizibleExpenses = vizibleExpenses
+        .filter(expense => expense.category === category);
+    }
+
+    if (from) {
+      vizibleExpenses = vizibleExpenses
+        .filter(expense => expense.spentAt >= from && expense.spentAt <= to);
+    }
+
+    res.send(vizibleExpenses);
```
AlinaMatuschak, line 29: return status code too

src/controllers/expenses.js
```diff
@@ -0,0 +1,101 @@
...
+      res.sendStatus(400);
+
+      return;
+    }
+
+    if (!userServices.getOne(userId)) {
+      res.sendStatus(400);
+
+      return;
+    }
```
AlinaMatuschak, line 59: Maybe don’t need to call create method if userId isn’t valid?

src/controllers/expenses.js
```diff
@@ -0,0 +1,101 @@
...
+
+    res.statusCode = 201;
+    res.send(newExpense);
+  },
+  remove: (req, res) => {
+    const id = req.params.expenseId;
```
AlinaMatuschak, line 65:
```suggestion
    const { expenseId } = req.params;
```

src/controllers/expenses.js
```diff
@@ -0,0 +1,101 @@
...
+      res.sendStatus(404);
+
+      return;
+    }
+
+    res.statusCode = 204;
+    res.send(filteredExpenses);
+  },
+  update: (req, res) => {
+    const expenseId = req.params.expenseId;
```
AlinaMatuschak, line 79:
```suggestion
    const { expenseId } = req.params;
```

src/controllers/users.js
```diff
@@ -0,0 +1,70 @@
+'use strict';
+
+const { userServices } = { ...require('../services/users') };
```
AlinaMatuschak, line 3:
```suggestion
const { userServices } = require('../services/users');
```

src/controllers/users.js
```diff
@@ -0,0 +1,70 @@
...
+
+const userController = {
+  getAll: (req, res) => {
+    const users = userServices.getAll();
+
+    res.send(users);
```
AlinaMatuschak, line 9: add status code for response

src/controllers/users.js
```diff
@@ -0,0 +1,70 @@
...
+  },
+  getOne: (req, res) => {
+    const userId = req.params.userId;
```
AlinaMatuschak, line 12:
```suggestion
    const { userId } = req.params;
```

src/createServer.js
```diff
@@ -1,11 +1,23 @@
 'use strict';
 
-// const express = require('express');
+const { expenseController } = { ...require('./controllers/expenses') };
+const { userController } = { ...require('./controllers/users.js') };
+const { usersRouter } = { ...require('./routes/users.js') };
+const { expensesRouter } = { ...require('./routes/expenses.js') };
```
AlinaMatuschak, line 6: 🥲

src/createServer.js
```diff
@@ -1,11 +1,23 @@
...
+const express = require('express');
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  app.use(cors());
+  app.use(usersRouter);
```
AlinaMatuschak, line 14: you can add here endpoint for user routes (you can delete `/users` from the user routes file)
```suggestion
  app.use('/users', usersRouter);
```

src/services/expenses.js
```diff
@@ -0,0 +1,57 @@
...
+
+    const newExpense = {
+      userId,
+      spentAt,
+      title,
+      amount,
+      category,
+      note,
+      id: newExpensesId++,
+    };
```
AlinaMatuschak, line 27:
```suggestion
    const newExpense = {
      ...body,
      id: newExpensesId++,
    };
```

AlinaMatuschak (changes requested):
Good start 👍. Just small issues )

src/controllers/expenses.js
```diff
@@ -0,0 +1,105 @@
...
+
+    if (category) {
+      vizibleExpenses = vizibleExpenses
+        .filter(expense => expense.category === category);
+    }
+
+    if (from) {
+      vizibleExpenses = vizibleExpenses
+        .filter(expense => expense.spentAt >= from && expense.spentAt <= to);
+    }
```
dellachaise, line 25: Maybe code would be more elegant in you would split this logic into 3 different methods. What do you think?

src/controllers/expenses.js
```diff
@@ -0,0 +1,105 @@
...
+
+    res.statusCode = 200;
+    res.send(vizibleExpenses);
+  }
+
+  getOne(req, res) {
+    const id = req.params.expenseId;
```
dellachaise, line 32: Still need to change(the same comment in the previous code review).
```suggestion
    const { expenseId } = req.params;
```

src/controllers/expenses.js
dellachaise: Here the same.

src/controllers/users.js
dellachaise: Here too.

tests/expense.test.js
```diff
@@ -188,6 +188,7 @@ describe('Expense', () => {
         });
 
       const response = await api
+        // eslint-disable-next-line max-len
```

dellachaise (approved):
Good job 👍 

Some of the comments from the previous code review you didn't fix and I see no conversation about that points. Be attentive, please.

PR #32, opened 2022-10-23
Node accounting app

PastolNapas:


OleksiiPustovit (approved):
Nice start, but it would be really cool if you split the main file))

src/createServer.js
```diff
@@ -1,11 +1,280 @@
 'use strict';
 
-// const express = require('express');
+const express = require('express');
 
 function createServer() {
   // Use express to create a server
   // Add a routes to the server
   // Return the server (express app)
```
dellachaise, line 8: Delete comments all over the project. And do not leave such comments in your future test tasks.

src/createServer.js
```diff
@@ -1,11 +1,280 @@
...
+
+    const findUser = users.find((user) => user.id === Number(id));
+
+    if (!findUser) {
+      res.statusCode = 404;
+
+      res.send();
+
+      return;
+    }
```
dellachaise, line 88: Do you really need `send` here? Why you can't write just this way? 👇 
```suggestion
      return;
    }
```

src/createServer.js
```diff
@@ -1,11 +1,280 @@
...
+
+    if (Number.isNaN(Number(id))) {
+      res.statusCode = 400;
+
+      res.send();
+
+      return;
```
dellachaise, line 95: Here the same, and the same in such cases all over the code.

src/createServer.js
```diff
@@ -1,11 +1,280 @@
...
+    }
+
+    if (from && to) {
+      const filteredExpenses = expenses.filter(
+        (expense) => expense.spentAt >= from && expense.spentAt <= to
+      );
+
+      res.statusCode = 200;
```
dellachaise, line 177: Maybe code would be more elegant in you would split this logic into 3 different methods. What do you think?

src/createServer.js
```diff
@@ -1,11 +1,280 @@
...
+      res.send();
+
+      return;
+    }
+
+    expenses.splice(0, expenses.length, ...filterExpense);
+
+    res.statusCode = 204;
+
+    res.send();
```
dellachaise, line 252: No need in send here, delete it.

src/createServer.js
```diff
@@ -1,11 +1,280 @@
...
+      res.statusCode = 404;
+
+      res.send();
+
+      return;
+    }
+
+    expenses.splice(0, expenses.length, ...filterExpense);
+
+    res.statusCode = 204;
```
dellachaise, line 250: use `sendStatus` here.

src/createServer.js
```diff
@@ -1,11 +1,280 @@
...
+    res.statusCode = 200;
+
+    res.send(users);
+  });
+
+  app.post('/users', express.json(), (req, res) => {
+    const { name } = req.body;
+
+    if (!name) {
+      res.statusCode = 400;
```
dellachaise, line 32: use `sendStatus` here. Do the same all over this task code and in other projects if you would need it.

tests/expense.test.js
```diff
@@ -188,6 +188,7 @@ describe('Expense', () => {
         });
 
       const response = await api
+      // eslint-disable-next-line max-len
```

dellachaise (changes requested):
Good start 👍
Create routes, controllers and services (split your code according to [MVC](https://mate.academy/learn/nodejs/express-nodejs#/video/2723))

src/controllers/expenses.js
```diff
@@ -0,0 +1,147 @@
...
+
+const addExpense = (req, res) => {
+  const { userId } = req.body;
+
+  const foundUser = userServise.getUserById(userId);
+
+  if (!foundUser) {
+    res.statusCode = 400;
+
+    res.send();
```
dellachaise, line 66: Still actual: for what purpose do you use `res.send();` here? 
Use `res.sendStatus(404);` in line 64 and then you could delete line 66.

Do the same all over the code in such cases.

src/controllers/expenses.js
```diff
@@ -0,0 +1,147 @@
...
+
+const getExpenseById = (req, res) => {
+  const expenseId = Number(req.params.id);
+
+  const foundExpense = expenseServise.getExpenseById(expenseId);
+
+  if (!foundExpense) {
+    res.statusCode = 404;
+
+    res.send();
```
dellachaise, line 91: Here the same. And check and fix the in the same cases all over the code 👇

src/controllers/expenses.js
```diff
@@ -0,0 +1,147 @@
...
+
+    return;
+  }
+
+  Object.assign(foundExpense, req.body);
+
+  res.statusCode = 200;
+
+  res.send(foundExpense);
```
dellachaise, line 138: Catch an error. What if something went wrong with removing?
 
The same better do on remove too.

dellachaise (approved):
Good job 👍 

Just add improvements according to the comments 👇

tests/expense.test.js
dellachaise: And you missed this comment ☝️  It is still actual.

PR #33, opened 2022-10-23
Solution

SvyatoslavYatsenko:


roma-shafihullin (approved):
good work! you can also split your app

src/createServer.js
```diff
@@ -1,12 +1,159 @@
+/* eslint-disable no-console */
```
AlinaMatuschak, line 1: Don’t disable eslint

src/createServer.js
```diff
@@ -1,12 +1,159 @@
...
+    if (!wantedUser) {
+      res.sendStatus(404);
+    };
+
+    res.json(wantedUser);
+  });
+
+  app.post('/users', (req, res) => {
+    const newUser = {
+      id: ++userId,
```
AlinaMatuschak, line 30: you can get random number

src/createServer.js
```diff
@@ -1,12 +1,159 @@
...
+      name: req.body.name,
+    };
+
+    if (!req.body.name || (users.find(user => user.name === req.body.name))) {
```
AlinaMatuschak, line 34: use destructuring

AlinaMatuschak, line 34: create a variable for the found user

src/createServer.js
```diff
@@ -1,12 +1,159 @@
...
+    };
+
+    res.sendStatus(404);
+  });
+
+  /* expenses */
+  const epxenses = [];
+  let expensesId = 0;
+
+  app.get('/expenses', (req, res) => {
```
AlinaMatuschak, line 74: don't forget about destructuring

src/createServer.js
```diff
@@ -1,12 +1,159 @@
...
+    if (req.query.category) {
+      const expensByCategory = epxenses
+        .find(expens => expens.category === req.query.category);
+
+      res.status(200).json([expensByCategory]);
```
AlinaMatuschak, line 79:
```suggestion
      res.statusCode = 200;
      res.send(expensByCategory);
```

src/createServer.js
```diff
@@ -1,12 +1,159 @@
...
+  });
+
+  /* expenses */
+  const epxenses = [];
+  let expensesId = 0;
+
+  app.get('/expenses', (req, res) => {
+    if (req.query.category) {
+      const expensByCategory = epxenses
+        .find(expens => expens.category === req.query.category);
```
AlinaMatuschak, line 77: Maybe better use the filter?

AlinaMatuschak (changes requested):
Good start 👍. Create routes, controllers and services )

src/createServer.js
SvyatoslavYatsenko: There is no actual difference between res.send and res.json but res.json has an extra option: if we want to send non-object data, res.json would convert it to json format.

PR #35, opened 2022-10-23
Express solution

wfurs0w:


Reptiloid044 (approved):
Good Job!

src/routes/users.js
```diff
@@ -0,0 +1,95 @@
...
+
+function createUsers(app) {
+  app.use('/users', router);
+  init();
+
+  router.get('/', (req, res) => {
+    const users = getAll();
+
+    res.send(users);
+    res.statusCode = 200;
```
ArturErmolenko, line 24: it will be great to transfer logic from router, in controllers, you could create function
```
function getAll(req, res) {
   const users = usersServise.getAll();

   res.statusCode = 200;
   res.send(users);
 }
```
and your route will look like this
`router.get('/', usersControllers.getAll);`

you also need to change other routes

src/routes/expenses.js
```diff
@@ -0,0 +1,139 @@
...
+  });
+
+  router.get('/:expenseId', (req, res) => {
+    const { expenseId } = req.params;
+
+    const foundExpens = getExpenseById(expenseId);
+
+    if (!foundExpens) {
+      res.sendStatus(404);
+
```
ArturErmolenko, line 48: you should add conditions for status 400

ArturErmolenko (changes requested):
Good job 👍 

Let's try to connect controllers file for users and expenses
also check conditions when you should add  status 400

PR #36, opened 2022-10-23
Develop

roma-shafihullin:


SvyatoslavYatsenko (approved):
GJ

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  if (from && to) {
+    const foundExpenses = expenses.filter(
+      (expense) => expense.spentAt >= from && expense.spentAt <= to);
+
+    res.send(foundExpenses);
+
+    return;
+  }
```
dellachaise, line 51: Maybe code would be more elegant in you would split this logic into 2 different methods. What do you think?

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,141 @@
...
+
+const createExpense = (req, res) => {
+  const {
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
```
dellachaise, line 64: You could destructure only userId here
```suggestion
    const { userId } = req.body;
```
As I understand, you need only it. And in `expenseWitId` use `...req.body`.

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,141 @@
...
+
+  if (!expenseTarget) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  expenses = expenses.filter(expense => expense.id !== +id);
+
+  res.sendStatus(204);
```
dellachaise, line 115: Catch an error. What if something went wrong with removing?

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,141 @@
...
+
+  if (!expenseTarget) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const updatedExpense = Object.assign(expenseTarget, req.body);
+
+  res.send(updatedExpense);
```
dellachaise, line 131: Here the same.

tests/expense.test.js
```diff
@@ -188,6 +188,7 @@ describe('Expense', () => {
         });
 
       const response = await api
+        // eslint-disable-next-line max-len
```

dellachaise (changes requested):
Great start 👍
Just create services too (split your code according to [MVC](https://mate.academy/learn/nodejs/express-nodejs#/video/2723))

PR #38, opened 2022-10-23
Develop

Anastasiia-Khmelovska:


vladgusakovskiy (approved):
Great 👍

ArturErmolenko (changes requested):
Good job 👍

```
function getAll() {
   return users;
 }
```
and function on you controllers file

```
function getAll(req, res) {
   res.status(200)
   res.send(users);
 }
```
than your route will look like this

ArturErmolenko (approved):
Well done 🚀

PR #39, opened 2022-10-24
Node accounting app

Reptiloid044:
Hi!
This pull request is still not ready to check, I will delete this notification when it will be ready.


src/controller/expenseController.js
```diff
@@ -0,0 +1,141 @@
...
+
+  if (typeof +userId !== 'number') {
+    res.sendStatus(400);
+
+    return;
+  };
+
+  if (expenses.length === 0) {
+    res.send([]);
+  };
```
dellachaise, line 45: Why did you do this check?( I asked, because just interested about purpose, not saying that it is wrong, just trying to understand)

src/controller/expenseController.js
```diff
@@ -0,0 +1,141 @@
...
+  const foundExpense = expenseServise.findById(+id);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  };
+
+  expenseServise.remove(+id);
+  res.sendStatus(204);
```
dellachaise, line 132: Here the same.

src/controller/userController.js
```diff
@@ -0,0 +1,96 @@
...
+  const foundUser = userServise.findById(+id);
+
+  if (!foundUser) {
+    res.sendStatus(404);
+
+    return;
+  };
+
+  if (typeof +id !== 'number') {
+    res.sendStatus(422);
```
dellachaise, line 67: I think `400` would be better in this case. Because 422 means:
 **the server understands the content type of the request entity**, and the syntax of the request entity **is correct**, but it was unable to process the contained instructions
 
 What do you think about it?

dellachaise (approved):
Good job 👍

PR #40, opened 2022-10-24
Develop

ibondrk:


RomanYakm (approved):
Good solution! 🚀 🔥

ArturErmolenko (changes requested):
Good job 👍 
Lets try to separate logic in different files (services & controllers & route)

PR #42, opened 2022-10-24
Solution

DimaKobzar7:


src/services/users.js
```diff
@@ -0,0 +1,49 @@
...
+
+function getUserById(userId) {
+  const foundUser = users.find(user => user.id === +userId);
+
+  return foundUser || null;
+}
+
+function createUser(name) {
+  const newUser = {
+    id: Math.max(0, ...users.map(({ id }) => id + 1)),
```
Desertiks, line 17: I'm not sure, but if the users array is very large, we will use too many resources.
I think better to create variable like `nextCounter = 1` and increse it every time we create user.

Desertiks (approved):
Great work 🔥

ArturErmolenko (approved):
Well done 👍

PR #43, opened 2022-10-24
Solution

Roman-Artiukh:


Andrii-Dunets (approved):
Good job)

src/createServer.js
```diff
@@ -1,11 +1,229 @@
 'use strict';
+// const { v4: uuidv4 } = require('uuid');
```
ArturErmolenko, line 2: delete comments

ArturErmolenko (changes requested):
Good job 🔥 
Let's try to separate this logic in different files (services & controllers& routes)

```
function getAll() {
   return users;
 }
```

```
function getAll(req, res) {
   res.status(200)
   res.send(users);
 }
```

PR #44, opened 2022-10-24
solutionNodeAccountingAppV2

YuriiPronin:


Maxim-Gumeniuk (approved):

src/createServer.js
```diff
@@ -1,11 +1,211 @@
...
+    const filtereedUser = users.filter((user) => user.id !== +id);
+
+    if (filtereedUser.length === users.length) {
+      res.sendStatus(404);
+
+      return;
+    }
+
+    users = filtereedUser;
+    res.sendStatus(204);
```
dellachaise, line 64: Catch an error. What if something went wrong with removing?

src/createServer.js
```diff
@@ -1,11 +1,211 @@
...
+    if (!findUser) {
+      res.sendStatus(404);
+    }
+
+    const { name } = req.body;
+
+    Object.assign(findUser, { name });
+
+    res.send(findUser);
+    res.statusCode = 200;
```
dellachaise, line 81: Here the same.

src/createServer.js
```diff
@@ -1,11 +1,211 @@
...
+
+  app.post('/expenses', (req, res) => {
+    const {
+      userId,
+      spentAt,
+      title,
+      amount,
+      category,
+      note,
+    } = req.body;
```
dellachaise, line 92: You could destructure only userId here
```suggestion
    const { userId } = req.body;
```
As I understand, you need only it. And in `newExpenses` use `...req.body`.

src/createServer.js
```diff
@@ -1,11 +1,211 @@
...
+    }
+
+    if (userId) {
+      const findUser = expenses.filter((expense) => expense.userId === +userId);
+
+      res.send(findUser);
+      res.statusCode = 200;
+
+      return;
+    }
```
dellachaise, line 151: Maybe code would be more elegant in you would split this logic into 3 different methods. What do you think?

PR #46, opened 2022-10-24
solution in one file

Ohiiko:


Andrii-Dunets (approved):
Good job)

src/controllers/expenses.js
```diff
@@ -0,0 +1,121 @@
...
+  const foundExpense = getByExpenseId(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  removeExpense(expenseId);
+  res.sendStatus(204);
```
dellachaise, line 97: Catch an error. What if something went wrong with removing?

src/services/expenses.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+let expenses = [];
+let newExpenceId = 1;
```
dellachaise, line 4: Check what ids you would have. Better use Math.random multiply with new Date inside of the function where you use it.

src/services/users.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+let users = [];
+let newUserId = 1;
```
dellachaise, line 4: The same.

dellachaise (approved):
Good job 👍

PR #50, opened 2022-10-24
Develop

Andrii-Dunets:
Поки немає роутера, і не розділено на файли.


Roman-Artiukh (approved):
GJ)

ArturErmolenko (changes requested):
Good job 👍 

```
function getAll() {
   return users;
 }
```

```
function getAll(req, res) {
   res.status(200)
   res.send(users);
 }
```

PR #54, opened 2022-10-26
Solution

Anatolii-Bukoros:


OleksiiPustovit (approved):
Nice start 👍
But it would be really cool if you split the main file.
Create routes, controllers and services/

src/createServer.js
```diff
@@ -1,11 +1,219 @@
...
+
+  app.post('/expenses', (req, res) => {
+    const { userId, spentAt, title, amount, category, note } = req.body;
+
+    if (!userId) {
+      return res.status(400).send('User ID is required');
+    }
+
+    if (!title) {
+      return res.status(400).send('Title is required');
```
ArturErmolenko, line 88: Do not rely on the unknown string, make constants for this.
const Errors = {
  ....
};

ArturErmolenko (changes requested):
Good job 👍

```
function getAll() {
   return users;
 }
```
```

function getAll(req, res) {
   res.status(200)
   res.send(users);
 }
```

PR #61, opened 2022-11-21
solution

ihorandrianov:


src/app.ts
```diff
@@ -0,0 +1,39 @@
...
+  public app: express.Application;
+  public port: number;
+
+  constructor(controllers, port) {
+    this.app = express();
+    this.port = port;
+
+    this.initializeMiddlewares();
+    this.initializeControllers(controllers);
+    this.initializeErrorHandling();
```
Andr1yK, line 15: u can create `init` function😃

src/db/db.instance.ts
```diff
@@ -0,0 +1,62 @@
...
+  private idCounter: number;
+
+  constructor(items = []) {
+    this.items = items;
+    this.idCounter = 0;
+  }
+
+  private getItemIndex(id: number): number {
+    const index = this.items.findIndex((item) => item.id === id);
+    return index;
```
Andr1yK, line 12:
```suggestion
    return this.items.findIndex((item) => item.id === id);
```

src/db/db.instance.ts
```diff
@@ -0,0 +1,62 @@
+class DbInstance {
+  private items: any[];
+  private idCounter: number;
+
+  constructor(items = []) {
+    this.items = items;
+    this.idCounter = 0;
+  }
+
+  private getItemIndex(id: number): number {
```
Andr1yK, line 10:
```suggestion
  private getIndex(id: number): number {
```

src/db/db.instance.ts
```diff
@@ -0,0 +1,62 @@
...
+    const index = this.items.findIndex((item) => item.id === id);
+    return index;
+  }
+
+  public getAll() {
+    return this.items;
+  }
+
+  public createNew<T>(payload: T) {
```
Andr1yK, line 19: As I think, we always create a new instance :)

```suggestion
  public create<T>(payload: T) {
```

src/expanses/expanses.controller.ts
```diff
@@ -0,0 +1,98 @@
...
+class ExpansesController {
+  public path = "/expanses";
+  public router = express.Router();
+
+  private expanses;
+
+  constructor() {
+    this.initializeRoutes();
+    this.expanses = new DbInstance();
+  }
```
Andr1yK, line 17: Let's create a basic (abstract) class `Controller` (you can add `initializeRoutes` method to it as required)

image

src/expanses/expanses.interface.ts
```diff
@@ -0,0 +1,11 @@
+interface Expanse {
```
Andr1yK, line 1: maybe u can use `CreateExpanseDto` as a type🤔

src/middleware/error.middleware.ts
```diff
@@ -0,0 +1,19 @@
+import { NextFunction, Request, Response } from "express";
+import HttpException from "exceptions/HttpException";
+
+function errorMiddleware(
+  error: HttpException,
+  _request: Request,
+  response: Response,
+  _next: NextFunction
```
Andr1yK, line 8: do u need `next` here?)

src/middleware/validation.middleware.ts
```diff
@@ -0,0 +1,28 @@
...
+import { validate, ValidationError } from "class-validator";
+import { plainToClass } from "class-transformer";
+import HttpException from "../exceptions/HttpException";
+
+function validationMiddleware(
+  type: any,
+  skipMissingProperties = false
+): RequestHandler {
+  return (req, _res, next) => {
+    validate(plainToClass(type, req.body), {
```
Andr1yK, line 11: I would use `async wait` :)

src/users/users.controller.ts
```diff
@@ -0,0 +1,97 @@
...
+    const id = request.params.id;
+    const user = this.users.getById(Number(id));
+    if (user) {
+      response.send(user);
+      return;
+    }
+    next(new UserNotFoundException(id));
+  };
+
+  private editUser = (
```
Andr1yK, line 65: as for the method, it would be more correct to call it `update`

Andr1yK (changes requested):

src/expanses/expanses.interface.ts
ihorandrianov: CreateExpanse should be without **id** (will fix it in next commit) so i can validate incoming JSON to not contain any external keys, and to not allow duplicate id which is generated on "server"

ihorandrianov: Same for User interface and User Dto

Andr1yK: u can use something looks like this 

```
type Expanse = CreateExpanseDto & {
  id: number
};

Comment (Andr1yK) [reply src/expanses/expanses.interface.ts:1]: or create a utility for this issue

type Expanse = ModelWithId<CreateExpanseDto>

File: src/users/users.controller.ts [hunk 10]
```diff
@@ -5,18 +5,20 @@ import validationMiddleware from "../middleware/validation.middleware";
...
-  private users;
+  private users: DbInstance;
 
   constructor() {
+    super();
     this.initializeRoutes();
     this.users = new DbInstance();
   }
 
-  private initializeRoutes() {
```
Andr1yK: u can make it private :)

src/users/users.interface.ts
```diff
@@ -1,6 +1,7 @@
-interface User {
+import CreateUserDto from "./users.dto";
+
+interface User extends CreateUserDto {
```
Andr1yK, line 3: :fire:

Andr1yK (approved):

src/users/users.controller.ts
ihorandrianov: abstract function cant be private compiler says

PR #64, opened 2022-11-22
Accounting app

Vladosik1337:


src/controllers/expenses.js
```diff
@@ -0,0 +1,107 @@
...
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
+
+  const user = userServices.getById(userId);
+
+  const isCorrect = user && spentAt && title && amount && category;
```
anastasiia-nahaiska, line 43: Maybe would be better to process a case if some fields won't be in the target type or a user with this userId doesn't exist

src/createServer.js
```diff
@@ -1,11 +1,17 @@
...
+
+const { router: usersRouter } = require('./routes/users');
+const { router: expensesRouter } = require('./routes/expenses');
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
```
anastasiia-nahaiska, line 10: Does your code work without cors?

anastasiia-nahaiska (approved):
GJ! 🔥

silchyn (approved):
Cool!

PR #65, opened 2022-11-22
Develop

AntonVaida:


src/services/expenses.js
```diff
@@ -0,0 +1,90 @@
...
+  {
+    id: 0,
+    userId: 1,
+    spentAt: '2022-11-21T13:35:39.907Z',
+    title: 'Phone',
+    amount: 0,
+    category: 'Electronics',
+    note: 'Buy new phone',
+  },
+];
```
artem-musii, line 13: expenses and users arrays should be empty at the beginning

src/routes/expenses.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const express = require('express');
+
+const expenseControllers = require('../controllers/expense');
+
+const router = express.Router();
+
+router.get('/expenses', express.json(), expenseControllers.getAllExpenses);
```
artem-musii, line 9: I think it's better to add app.use(express.json()) in createServer to avoid repeating of code

artem-musii (approved):
GJ!

src/controllers/expense.js
```diff
@@ -0,0 +1,113 @@
...
+const getExpense = (req, res) => {
+  const { expenseId } = req.params;
+
+  if (!expenseServices.getExpense(expenseId)) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(expenseServices.getExpense(expenseId));
```
Denchepornyuk, line 39: move `expenseServices.getExpense(expenseId)` to a separate variable

src/controllers/expense.js
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  res.send(status);
+};
+
+module.exports.getAllExpenses = getAllExpenses;
+module.exports.getExpense = getExpense;
+module.exports.deleteExpense = deleteExpense;
+module.exports.createNewExpense = createNewExpense;
+module.exports.updateExpense = updateExpense;
```
Denchepornyuk, line 113: you can write it all in module.exports object, fix for all controllers
```
{
  module.exports = {
    getAllExpenses,
    getExpense,
    ...
  }
}
```

src/controllers/user.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
+
+const userServices = require('../services/users');
+
+const getAll = (req, res) => {
+  res.send(userServices.getAllUsers());
```
Denchepornyuk, line 6: move `userServices.getAllUsers()` to a separate variable

src/controllers/user.js
```diff
@@ -0,0 +1,71 @@
...
+const getUser = (req, res) => {
+  const { todoId } = req.params;
+
+  if (!userServices.getUser(todoId)) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(userServices.getUser(todoId));
```
Denchepornyuk, line 18: the same

src/routes/expenses.js
Denchepornyuk: agree, move express.json() to server

Denchepornyuk, line 9: add base path `/expenses` also in server, fix for all routes

src/services/expenses.js
```diff
@@ -0,0 +1,80 @@
...
+  });
+
+  return findExpenses;
+}
+
+module.exports.getAllExpenses = getAllExpenses;
+module.exports.getExpense = getExpense;
+module.exports.deleteExpense = deleteExpense;
+module.exports.createNewExpense = createNewExpense;
+module.exports.updateExpense = updateExpense;
```
Denchepornyuk, line 80: make your exports simpler

src/services/expenses.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+function createNewExpense(userId, spentAt, title, amount, category, note) {
+  let newId = 0;
+
+  if (expenses.length) {
+    newId = [...expenses].sort(
+      (expenseA, expenseB) => expenseB.id - expenseA.id
+    )[0].id + 1;
+  }
```
Denchepornyuk, line 43: this code is difficult to understand, don't try to write code in a couple of lines, create additional variables

src/services/users.js
```diff
@@ -0,0 +1,60 @@
...
+  return users.find(user => user.id === Number(userId));
+}
+
+function createNewUser(name) {
+  let newId = 0;
+
+  if (users.length) {
+    newId = [...users].sort(
+      (userA, userB) => userB.id - userA.id
+    )[0].id + 1;
```
Denchepornyuk, line 19: you can create folder utils and move repeated code there, for finding correct id you can create a function

Denchepornyuk (changes requested):

src/controllers/expense.js
```diff
@@ -0,0 +1,121 @@
...
+  updateExpense,
+} = require('../services/expenses');
+
+const getAllExpensesController = (req, res) => {
+  const { userId, category, from, to } = req.query;
+
+  if (!category || !userId || !from || !to) {
+    res.sendStatus(422);
+
+    return;
```
OleksiiBlezniuk, line 17: Why do we need to return 422 error here? It's a method for getting all todos, and it should work even if no filter params is passed

src/controllers/expense.js
```diff
@@ -0,0 +1,121 @@
...
+const getExpenseController = (req, res) => {
+  const { expenseId } = req.params;
+
+  if (!getExpense(expenseId)) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(getExpense(expenseId));
```
OleksiiBlezniuk, line 45: You've called this function twice, just save it's result to constant and use this value

src/controllers/expense.js
```diff
@@ -0,0 +1,121 @@
...
+};
+
+const deleteExpenseController = (req, res) => {
+  const { expenseId } = req.params;
+
+  const status = deleteExpense(expenseId);
```
OleksiiBlezniuk, line 51: it's not clear why status has boolean value

better to call it `wasRemoved` or smth like that

src/controllers/expense.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  const newExpense = createNewExpense(
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note
+  );
```
OleksiiBlezniuk, line 81: Why not just is as one object, or even send body? You validated this values and sure that they are valid, nothing is happened with this values inside `createNewExpense` function, I don't see any reason to pass every object param as separate argument

src/controllers/expense.js
```diff
@@ -0,0 +1,121 @@
...
+    return;
+  }
+
+  const status = updateExpense(
+    expenseId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note
```
OleksiiBlezniuk, line 103: Same here, only id is used as separate constant inside `updateExpense` function, you can pass object here

src/controllers/user.js
```diff
@@ -0,0 +1,79 @@
...
+const getUserController = (req, res) => {
+  const { todoId } = req.params;
+
+  if (!getUser(todoId)) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(getUser(todoId));
```
OleksiiBlezniuk, line 24: Save to constant, why calling this function twice?

src/controllers/user.js
```diff
@@ -0,0 +1,79 @@
...
+const deleteUserController = (req, res) => {
+  const { userId } = req.params;
+
+  const status = deleteUser(userId);
+
+  if (status) {
+    res.sendStatus(204);
+  } else {
+    res.sendStatus(404);
+  }
```
OleksiiBlezniuk, line 51:
```suggestion
  if (!status) {
    res.sendStatus(404);
    
    return;
  }
  
    res.sendStatus(204);
```

src/routes/expenses.js
```diff
@@ -0,0 +1,26 @@
...
+
+router.get('/', getAllExpensesController);
+
+router.get('/:expenseId', getExpenseController);
+
+router.delete('/:expenseId', deleteExpenseController);
+
+router.post('/', createNewExpenseController);
+
+// eslint-disable-next-line
```
OleksiiBlezniuk, line 23: why do you need to disable linter here?

src/services/expenses.js
```diff
@@ -0,0 +1,82 @@
...
+  return true;
+}
+
+function getExpense(expenseId) {
+  return expenses.find(
+    expense => expense.id === Number(expenseId)
+  );
+}
+
+function createNewExpense(userId, spentAt, title, amount, category, note) {
```
OleksiiBlezniuk, line 38: Better pass object here

src/services/expenses.js
```diff
@@ -0,0 +1,82 @@
...
+    category,
+    note,
+  };
+
+  expenses.push(newExpense);
+
+  return newExpense;
+}
+
+function updateExpense(expenseId, spentAt, title, amount, category, note) {
```
OleksiiBlezniuk, line 60: same here

OleksiiBlezniuk (changes requested):
GJ, but avoid unnecessary function calls

OleksiiBlezniuk (approved):
GJ!

PR #66, opened 2022-11-22
Expense tracking app has been implemented

iam-datsenko:


src/Users/usersControllers.js
```diff
@@ -0,0 +1,90 @@
...
+
+  const foundUser = userServices.getById(userId);
+
+  if (!foundUser) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(foundUser);
```
imaksweb, line 26:
```suggestion
  res.statusCode = 200;
  res.send(foundUser);
```

src/Users/usersControllers.js
```diff
@@ -0,0 +1,90 @@
...
+
+    return;
+  }
+
+  userServices.update({
+    id: userId,
+    name,
+  });
+
+  res.send(foundUser);
```
imaksweb, line 81:
```suggestion
  res.statusCode = 200;
  res.send(foundUser);
```

src/Expenses/expensesControllers.js
```diff
@@ -0,0 +1,111 @@
...
+};
+
+const getOneExpense = (req, res) => {
+  const { expenseId } = req.params;
+
+  if (!expenseId) {
+    res.sendStatus(400);
+
+    return;
+  }
```
imaksweb, line 16: I'm not sure you should do this check because if you don't have valid `expenseId` you won't find `foundExpense` and you  already handled this case further in the code

src/Expenses/expensesControllers.js
```diff
@@ -0,0 +1,111 @@
...
+
+  const foundExpense = expensesServices.getById(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(foundExpense);
```
imaksweb, line 26:
```suggestion
  res.statusCode = 200;
  res.send(foundExpense);
```

src/Expenses/expensesControllers.js
```diff
@@ -0,0 +1,111 @@
...
+  const {
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
+
+  const hasBody = userId && spentAt && title && amount && category && note;
```
imaksweb, line 39: According to documentation `note` is not obligatory field)

imaksweb (approved):
Nice work 👍 
Just left some comments 🔽

src/Users/usersControllers.js
iam-datsenko: Status 200 will be by default

src/Expenses/expensesControllers.js
iam-datsenko: It is tech requirements:
404 if the expected entity doesn't exist (if I have no id)
400 if the required parameter is not passed (if I have an id, but it isn't an appropriate one)

src/Expenses/expensesControllers.js
iam-datsenko: Yeah, thnx

src/Expenses/expensesServices.js
```diff
@@ -0,0 +1,69 @@
...
+
+const getAll = () => expenses;
+
+const getById = (expenseId) => {
+  return expenses.find(expense => expense.id === +expenseId);
+};
+
+const create = ({
+  userId,
+  spentAt,
```
Andr1yK, line 13: u can create class for service

Andr1yK (approved):

PR #68, opened 2022-11-22
Soultion

imaksweb:


src/controllers/expenses.js
```diff
@@ -0,0 +1,113 @@
...
+  const { expenseId } = req.params;
+  const foundExpense = expenseService.getById(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.statusCode = 200;
```
iam-datsenko, line 31: Status 200 will be by default, check it everywhere

src/controllers/expenses.js
```diff
@@ -0,0 +1,113 @@
...
+  const newExpense = expenseService.create(
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note
+  );
+
+  res.statusCode = 201;
```
iam-datsenko, line 67: Use method `res.status()` here

src/services/expenses.js
```diff
@@ -0,0 +1,77 @@
...
+let expenses = [];
+
+function getAll() {
+  return expenses;
+}
+
+function getById(expenseId) {
+  const foundExpense = expenses.find(expense => expense.id === +expenseId);
+
+  return foundExpense || null;
```
iam-datsenko, line 12: I think `null ` is extra here

iam-datsenko (approved):
GJ!

src/services/expenses.js
imaksweb: I'm not sure to be honest🙂

PR #69, opened 2022-11-23
Accounting App

Yuriy-Komolov:


Anna-Borodaienko (approved):

sergii-nosachenko (changes requested):
Fix linter error and pass all checks.
You can disable linter `max-len` rule for the problematic line in tests.

PR #71, opened 2022-11-23
Accounting map was implemented

DenysKleimenov:
I've implemented it with classes and **bind** was used for stable work of **this** in controllers

### DIFF
--- a/src/controllers/users.js
+++ b/src/controllers/users.js
@@ -1,0 +1,82 @@
+'use strict';
+
+const { Users } = require('../services/users.js');
+
+class UserController {
+  constructor() {
+    this.usersService = new Users();
+  }
+
+  getAll(req, res) {
+    const users = this.usersService.getAll();
+
+    res.send(users);
+  };
+
+  getOne(req, res) {
+    const { id } = req.params;
+    const foundUser = this.usersService.getUserById(id);
+
+    if (!foundUser) {
+      res.sendStatus(404);
+
+      return;
+    }
+
+    res.send(foundUser);
+  }
+
+  add(req, res) {
+    const { name } = req.body;
+
+    if (!name) {
+      res.sendStatus(400);
+
+      return;
+    }
+
+    const newUser = this.usersService.create(name);
+
+    res.status(201);
+    res.send(newUser);
+  }
+
+  change(req, res) {
+    const { id } = req.params;
+    const foundUser = this.usersService.getUserById(id);
+
+    if (!foundUser) {
+      res.sendStatus(404);
+
+      return;
+    }
+
+    const { name } = req.body;
+
+    if (typeof name !== 'string') {
+      res.sendStatus(422);
+
+      return;
+    }
+
+    this.usersService.update(id, name);
+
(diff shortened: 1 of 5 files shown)


ArturNagaycev (approved):
Cool! ⭐ 

I think, it would be better to add Routers for users and expenses

src/createServer.js
```diff
@@ -1,11 +1,54 @@
...
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  app.use(express.json());
+
+  const userController = new UserController();
+  const expensesController = new ExpensesController();
+
+  app.post('/users', userController.add.bind(userController));
```
sergii-nosachenko, line 15:
```suggestion
  app.post('/users', userController.add);
```
Better bind this inside constructor (I've added example)

But in ES6 class syntax you can define methods as arrow functions so this will always points to class itself.

src/controllers/users.js
```diff
@@ -0,0 +1,82 @@
+'use strict';
+
+const { Users } = require('../services/users.js');
+
+class UserController {
+  constructor() {
+    this.usersService = new Users();
```
sergii-nosachenko, line 7:
```suggestion
    this.usersService = new Users();
    this.add = this.add.bind(this);
```

src/createServer.js
```diff
@@ -1,11 +1,54 @@
...
+    if (!foundUser) {
+      res.sendStatus(400);
+
+      return;
+    }
+
+    const newExpense = expensesController.expensesService.create(req.body);
+
+    res.status(201);
+    res.send(newExpense);
```
sergii-nosachenko, line 38: You've created a bunch of problems when decided to use classes for your app. As for me the better approach is to have all methods as separate functions: no `this` problem, better reusability and flexibility. For example, this function could be moved to Expenses controller too.

src/services/expenses.js
```diff
@@ -0,0 +1,63 @@
...
+      }
+
+      if ((from && to) && (spentAt.localeCompare(from) === -1
+      || spentAt.localeCompare(to) === 1)) {
+        return false;
+      }
+
+      if (category && expense.category !== category) {
+        return false;
+      }
```
sergii-nosachenko, line 37: You need to check an array, not a single value.
Accounting-App-1-0-0-YURIIHOLIUK_1-SwaggerHub

src/services/expenses.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
+
+class Expenses {
+  constructor() {
+    this.expenses = [];
+  }
+
+  create(data) {
+    const maxId = Math.max(...this.expenses.map(user => user.id));
+    const id = (this.expenses.length > 0) ? maxId + 1 : 1;
```
sergii-nosachenko, line 10: You can create utility function for this and reuse it in all your services

sergii-nosachenko (changes requested):
Good code style. Just fix some issues.

sergii-nosachenko (approved):
Good work 👍

PR #74, opened 2022-11-23
simple accounting server was created

iPanda177:


src/createServer.js
```diff
@@ -1,11 +1,215 @@
 'use strict';
 
-// const express = require('express');
+const express = require('express');
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
```
iam-datsenko, line 6: Create separate files for services, routes and controllers (MVC)

src/createServer.js
```diff
@@ -1,11 +1,215 @@
...
+
+  let expenses = [];
+  let users = [];
+
+  app.use(express.json());
+
+  app.get('/users', (req, res) => {
+    res.statusCode = 200;
```
iam-datsenko, line 14: Status 200 will be by default

src/createServer.js
```diff
@@ -1,11 +1,215 @@
...
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  let expenses = [];
+  let users = [];
+
+  app.use(express.json());
```
iam-datsenko, line 11: if u add here '/users' or'/expenses' - so u don't have to write the whole path in each request

src/createServer.js
```diff
@@ -1,11 +1,215 @@
...
+      return;
+    }
+
+    const newUser = {
+      id: users.length + 1,
+      name,
+    };
+
+    users.push(newUser);
+    res.statusCode = 201;
```
iam-datsenko, line 53: use method res.status(201)

src/createServer.js
```diff
@@ -1,11 +1,215 @@
...
+
+    expenses = filteredExpense;
+    res.sendStatus(204);
+  });
+
+  app.patch('/expenses/:id', (req, res) => {
+    const { id } = req.params;
+    const findExpense = expenses.find((expense) => expense.id === +id);
+
+    if (typeof +id !== 'number') {
```
iam-datsenko, line 195: +string will be NaN, typeof NaN === 'number'

iam-datsenko (approved):
GJ!

OleksiiBlezniuk (changes requested):
Split your logic into separate files (services, routers, controllers)

PR #75, opened 2022-11-24
Develop

Oleksandra-Shevchenko:


Maksym845 (approved):
Good job! If you can, check my code, please [pr](https://github.com/mate-academy/node_accounting-app-v2/pull/77)

src/services/expenses.ts
```diff
@@ -0,0 +1,58 @@
+'use strict';
+
+let expenses: any[] = [];
+
+const getAllExpenses = () => {
+  return expenses;
+};
+
+const getExpenseById = (expenseId) => {
```
OleksiiBlezniuk, line 9: this file has TS extension, array of data has type, but function arguments are not typed (as well as function result)
Add types here, or change extension and write on JS, use one option 🙂

OleksiiBlezniuk (changes requested):
GJ, but use either JS or TS, not both at one time

OleksiiBlezniuk (commented):
Something wrong, now I can't see `controllers` folder 🥲 Probably you forgot to add this files to commit

src/services/expenses.js
```diff
@@ -0,0 +1,59 @@
...
+  Object.assign(foundExpense, {
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  });
+
+  return foundExpense;
+};
```
OleksiiBlezniuk, line 51:
```suggestion
const updateExpenses = ({
  id,
  ...expenseData,
}) => {
  const foundExpense = getExpenseById(id);

  Object.assign(foundExpense, expenseData);

  return foundExpense;
};
```

OleksiiBlezniuk: We need only id, other data can be just updated that way, no need to specify every param

OleksiiBlezniuk (approved):

PR #76, opened 2022-11-24
node accounting app commonjs on classes

TheVovchik:


illia-sosnytskyi (approved):
looks interesting how you created `routers` and decided not to use `send` and `sendStatus` from `express`

tests/expense.test.js
```diff
@@ -53,7 +54,7 @@ describe('Expense', () => {
 
     it('should return 400 if user not found', async() => {
       const expenseData = {
-        userId: 1,
```
sergii-nosachenko, line 56: Why do you change it?

sergii-nosachenko (changes requested):
You need to pass all tests before submitting PR. Try to commit all package files.

tests/expense.test.js
TheVovchik: try to understand why server not shutted down, return it to initial state now

src/controller/expensesController.js
```diff
@@ -0,0 +1,178 @@
...
+    const { userId, spentAt, title, amount, category, note } = req.body;
+
+    const users = usersController.getAll();
+    const hasUser = users.find(user => +userId === user.id);
+
+    const isValid = hasUser && spentAt && title && amount && category && note;
+
+    const error = hasUser
+      ? 'required parameters is not passed'
+      : 'user does\'nt exist';
```
sergii-nosachenko, line 32: You don't need error here, move it inside if block.

src/controller/expensesController.js
```diff
@@ -0,0 +1,178 @@
...
+        (expense) => expense.spentAt > from && expense.spentAt < to
+      );
+
+      res.statusCode = 200;
+      res.json(expensesByDate);
+
+      return;
+    }
+
+    if (userId && category) {
```
sergii-nosachenko, line 65: Category can be array, check api docs.

sergii-nosachenko, line 65: Filter by userId in separate block

src/controller/expensesController.js
```diff
@@ -0,0 +1,178 @@
...
+
+    if (from && to) {
+      const expensesByDate = expenses.filter(
+        (expense) => expense.spentAt > from && expense.spentAt < to
+      );
+
+      res.statusCode = 200;
+      res.json(expensesByDate);
+
+      return;
```
sergii-nosachenko, line 62: You need to filter expenses by multiple options, not only `from` & `to`. Combine all conditions and return filtered array of expenses at once.

src/controller/expensesController.js
```diff
@@ -0,0 +1,178 @@
...
+
+      res.statusCode = 200;
+      res.json(expensesByCategory);
+
+      return;
+    }
+
+    res.statusCode = 200;
+    res.json(expenses);
+  };
```
sergii-nosachenko, line 101: This method is incorrect - filter array by all available query params.

src/controller/usersController.js
```diff
@@ -0,0 +1,115 @@
+'use strict';
+
+const { UsersService } = require('../services/users');
+
+class UsersController extends UsersService {
```
sergii-nosachenko, line 5: Totally disagree.

src/controller/usersController.js
```diff
@@ -0,0 +1,115 @@
...
+    const data = super.getAll();
+
+    res.statusCode = 200;
+    res.json(data);
+  };
+
+  getUser(req, res) {
+    const { userId } = req.params;
+
+    if (isNaN(userId)) {
```
sergii-nosachenko, line 47:
```suggestion
    if (isNaN(+userId)) {
```

src/controller/usersController.js
```diff
@@ -0,0 +1,115 @@
...
+    res.sendStatus(204);
+  };
+
+  patchUser(req, res) {
+    const { userId } = req.params;
+    const { name } = req.body;
+
+    const error = !name
+      ? 'required parameter is not passed'
+      : 'required parameter is not valid, expected number';
```
sergii-nosachenko, line 88: Why do expect number? What parameter? Give detailed information about what is incorrect.

src/routes/expenses.js
```diff
@@ -0,0 +1,26 @@
...
+
+const express = require('express');
+
+// #region expensesRouter
+const expensesRouter = express.Router();
+
+function expences(app) {
+  app.use('/expenses', expensesRouter);
+  expensesController.setInitialExpenses();
+  expensesRouter.use(express.json());
```
sergii-nosachenko, line 15:
```suggestion
```
You can set global middleware in createServer.js as you did it with cors.

src/services/expenses.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+let expenses = [];
+
+class ExpensesService {
+  setInitialExpenses() {
+    expenses = [];
+  }
```
sergii-nosachenko, line 8: If you create your service as separate class, then you can store expenses inside it. Initialise em inside constructor.

src/services/expenses.js
```diff
@@ -0,0 +1,74 @@
...
+    return hasDeleted;
+  }
+
+  modifyExpence(expenceId, {
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  }) {
```
sergii-nosachenko, line 57:
```suggestion
  modifyExpence(expenceId, data) {
```

Don't need to destructurize. You can assign `data` as object. It can be partial, of course. Only available in data keys will be overwritten.

sergii-nosachenko (changes requested):

src/controller/expensesController.js
TheVovchik: it cann't

TheVovchik: I was using strict data schema from api docs, and here category was only string 
image

sergii-nosachenko: You are showing me `PATCH / expenses/:id` method, but I've commented you about `GET /expenses` method. Look closely. `Category` (not `categories` - it's incorrect) can be array. F.e.: 

```http
GET /expenses?userId=1&category=Food&category=Sport
```

You need to filter and include both categories. It's not present in tests but [described here](https://app.swaggerhub.com/apis/YURIIHOLIUK_1/Accounting-App/1.0.0#/Expenses/getExpenses).

src/routes/expenses.js
```diff
@@ -5,14 +5,14 @@ const {
...
+const { expensesService } = require('../services/expenses');
 
 // #region expensesRouter
 const expensesRouter = express.Router();
 
 function expences(app) {
   app.use('/expenses', expensesRouter);
-  expensesController.setInitialExpenses();
-  expensesRouter.use(express.json());
+  expensesService.setInitialExpenses();
```
sergii-nosachenko, line 15:
```suggestion
```

You don't need this method, cause every time you start server in calls `new expensesService()`. It call constructor, where you set empty array.

src/routes/users.js
```diff
@@ -5,14 +5,14 @@ const {
...
+const { usersService } = require('../services/users');
 
 // #region usersRouter
 const usersRouter = express.Router();
 
 function users(app) {
   app.use('/users', usersRouter);
-  usersController.setInitialUsers();
-  usersRouter.use(express.json());
+  usersService.setInitialUsers();
```
sergii-nosachenko, line 15:
```suggestion
```

sergii-nosachenko (changes requested):
Small fixes and you will be good to go.

src/routes/expenses.js
TheVovchik: it not works like you write and want. if we talk about server close/open it will work, but in tests. 
`beforeEach(() => {
    server = createServer();
    api = supertest(server);
  });`
  we just rerun function createServer(), so I need to create my objects with methods and with data inside this function call, or functions that I call in it.
  but Controllers method needs call service methods, so I need to make connection between them and in the same time they need to be in one function block to be recreated every time tests run different package of tests. 
  OR I make MVC separated files or I put them in one. 
  Once more test restart function - not server! so class methods will not be restarted, I need to recreate methods every time in routes! It's very stupid. We tell about this on classes, in QnA chanel. That's why I need here this function, so please also look logic of tests.
  
  createServer() => users() - routes + expenses() - routes - here I mention functions that restarts every test. and in expenses I need methods from user service....
  Every server start my code will create new instance of classes. But TESTS doesn't restart server, I cann't connect 4 separate class object and create them in one of routes function. I need to clean up data every test, I did it and I have no idea how to avoid clean up method

src/routes/users.js
TheVovchik: > Small fixes and you will be good to go.

look please on test logic! it rerun function not a Server! it's very important! I cann't transfer objects created in function between files and I need to connect Service and Controller logic! I can avoid reset of this in class object's but tests will not work!! 
once more 1) look on test logic! 2) It's not my decision to rewrite every this of data, I did it only to pass the test logic! 3) test logic must simulate RESTART of the server, not recall of the function! 4) I and many of student's, we told this to mentors, also we wrote this in qna chanel! Mentors told to skip test logic and tests because they not simulate server close/open.

PR #77, opened 2022-11-24
solution was added

Maksym845:
I have no idea why tests fail when I rewrite my server on functions, instead classes.


Oleksandra-Shevchenko (approved):
GJ 💯

src/services/expense.js
```diff
@@ -0,0 +1,77 @@
+'use strict';
+
+class Expense {
```
OleksiiBlezniuk, line 3: From `Expense` name I expect that I this is an object of Expense interface or it's class/constructor

OleksiiBlezniuk, line 3:
```suggestion
class ExpenseService {
```

src/services/users.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+class Users {
```
OleksiiBlezniuk, line 3:
```suggestion
class UsersService {
```

src/createServer.js
```diff
@@ -1,11 +1,144 @@
...
+
+    if (!user) {
+      res.sendStatus(404);
+
+      return;
+    }
+
+    res.statusCode = 200;
+    res.send(user);
+  });
```
OleksiiBlezniuk, line 40: Let's create routers for every entity and controllers for actions 🙂

src/createServer.js
```diff
@@ -1,11 +1,144 @@
...
+
+  app.delete('/users/:id', express.json(), (req, res) => {
+    const user = users.deleteUser(req.params.id);
+
+    if (user) {
+      res.statusCode = 204;
```
OleksiiBlezniuk, line 46: why not `sendStatus`?

src/createServer.js
```diff
@@ -1,11 +1,144 @@
...
+    }
+
+    res.sendStatus(404);
+  });
+
+  app.patch('/users/:id', express.json(), (req, res) => {
+    const user = users.updateUser(req.params.id, req.body);
+
+    if (user) {
+      res.statusCode = 200;
```
OleksiiBlezniuk, line 59: this status will be sent by default

src/createServer.js
```diff
@@ -1,11 +1,144 @@
...
+  });
+
+  app.post('/expenses', express.json(), (req, res) => {
+    const { title } = req.body;
+
+    if (!title) {
+      res.sendStatus(400);
+
+      return;
+    }
```
OleksiiBlezniuk, line 88: What about other properties?

src/createServer.js
```diff
@@ -1,11 +1,144 @@
...
+    const expense = expenses.updateExpense(req.params.id, req.body);
+
+    if (expense) {
+      res.statusCode = 200;
+      res.send(expense);
+
+      return;
+    }
+
+    res.sendStatus(404);
```
OleksiiBlezniuk, line 125: Better to validate data before updating data in storage, and not vice versa 🙂

src/services/expense.js
```diff
@@ -0,0 +1,77 @@
...
+  constructor() {
+    this.expenses = [];
+  }
+
+  getExpenses(query) {
+    if (query.userId && query.category) {
```
OleksiiBlezniuk, line 9: destructure query object

src/services/expense.js
```diff
@@ -0,0 +1,77 @@
...
+      return this.expenses.filter(expense => expense.userId === +query.userId
```
OleksiiBlezniuk, line 10: move conditions inside filter instead

src/services/expense.js
```diff
@@ -0,0 +1,77 @@
...
+      this.expenses.push(newExpense);
+
+      return newExpense;
+    }
+
+    return false;
+  }
+
+  getExpenseById(expenseId) {
+    return this.expenses.find(expense => expense.id === +expenseId);
```
OleksiiBlezniuk, line 52:
```suggestion
    return this.expenses.find(expense => expense.id === Number(expenseId));
```

src/services/expense.js
```diff
@@ -0,0 +1,77 @@
...
+    const foundExpense = this.expenses
+      .find(expense => expense.id === +expenseId);
+
+    if (foundExpense) {
+      Object.assign(foundExpense, newData);
+
+      return foundExpense;
+    }
+
+    return false;
```
OleksiiBlezniuk, line 65: It's strange that your function return object or boolean, in that case more clear to return null if action is not performed

src/services/users.js
```diff
@@ -0,0 +1,54 @@
...
+    let id = 0;
+
+    if (this.users.length !== 0) {
+      id = this.users[this.users.length - 1].id + 1;
+    }
+
+    const newUser = {
+      id,
+      name: newUserName,
+    };
```
OleksiiBlezniuk, line 22: Can id be `0`?

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

PR #79, opened 2022-11-24
server was implemented

ArturNagaycev:


src/controllers/expenses.js
```diff
@@ -0,0 +1,118 @@
...
+    note,
+  } = req.body;
+
+  if (typeof userId !== 'number'
+    || typeof spentAt !== 'string'
+    || typeof title !== 'string'
+    || typeof amount !== 'number'
+    || typeof category !== 'string'
+    || typeof note !== 'string'
+  ) {
```
DenysKleimenov, line 40: I think we can check if there is a user with a user Id that was passed in the request body and if not return status 400

src/controllers/expenses.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  const newExpense = expensesService.addNewExpense(
+    Number(userId),
+    spentAt,
+    title,
+    Number(amount),
+    category,
+    note,
+  );
```
DenysKleimenov, line 53: What's the point of using typecasting here if you have a checking of types above?

src/services/expenses.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+const addNewExpense = (
+  userId,
+  spentAt,
+  title,
+  amount,
+  category,
+  note,
+) => {
```
DenysKleimenov, line 22: You can just accept an object and then use the spread operator when creating a new expense.

DenysKleimenov (approved):
Looks good 🔥
Check some comments below

sergii-nosachenko (changes requested):
Fix linter error (just disable the `max-len` rule for problematic line in tests).

sergii-nosachenko (changes requested):
You need to pass all tests before requesting review.

PR #80, opened 2022-11-24
Accounting app (with Node.js)

Anna-Borodaienko:


Yuriy-Komolov (approved):
Excellent! Well done! 👍

src/controllers/expenses_controllers.js
```diff
@@ -0,0 +1,103 @@
+'use strict';
+
+const { expenseServices } = require('../services/expenses_services');
+const { userServices } = require('./users_controllers');
```
Anna-Tkachenko, line 4: it's not clear why services imported from controllers

Anna-Tkachenko (approved):
Good job! ⭐️

PR #82, opened 2022-11-25
implemented accounting-app

nikitaovch99:


vladyslav-harkusha (approved):
GJ

sergii-nosachenko (changes requested):
Fix linter error (just disable the `max-len` rule for problematic line in tests).

nikitaovch99:
> You need to pass all tests before review.

We discussed this with Denys Cheporniuk. For some reason, if you start doing MVC  in this task. Some tests won't pass. So he said better do MVC than trying to pass all tests.

src/controllers/expenses.js
```diff
@@ -0,0 +1,98 @@
...
+    return;
+  }
+
+  expensesServices.remove(expenseId);
+
+  res.sendStatus(204);
+};
+
+const update = (req, res) => {
+  const { title } = req.body;
```
sergii-nosachenko, line 68: Not only title

sergii-nosachenko (changes requested):

PR #89, opened 2022-11-28
Develop

myroslav-diiak:


src/createServer.js
```diff
@@ -1,11 +1,190 @@
 'use strict';
 
 // const express = require('express');
```
grafit-off, line 3: Useless comment

src/createServer.js
```diff
@@ -1,11 +1,190 @@
...
+const express = require('express');
+
+const { Expenses } = require('./services/Expenses.js');
+const { Users } = require('./services/Users.js');
 
 function createServer() {
   // Use express to create a server
   // Add a routes to the server
   // Return the server (express app)
```
grafit-off, line 12: I think u don't need these comments

src/createServer.js
```diff
@@ -1,11 +1,190 @@
...
+  const expenses = new Expenses();
+  const users = new Users();
+
+  app.use(express.json());
+
+  app.get('/users', (req, res) => {
+    res.send(users.getUsers());
+  });
+
+  app.get('/users/:userId', (req, res) => {
```
grafit-off, line 24: It's gonna be better to use Routes with prefixes, so we don't gonna have duplicates in endpoints like `/users` or `/expenses`

src/createServer.js
```diff
@@ -1,11 +1,190 @@
...
+  });
+
+  app.get('/expenses', (req, res) => {
+    const searchParams = req.query;
+
+    if (!searchParams) {
+      res.send(expenses.getExpenses());
+
+      return;
+    }
```
grafit-off, line 97: I'm not sure, but `req.query` it's an object, so that condition cant be falsy, isn't?

src/services/expenses.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  deleteExpense(expenseId) {
+    this.expenses = this.expenses.filter(expense => expense.id !== +expenseId);
+  }
+
+  updateExpense(expenseId, expenseBody) {
+    const foundExpense = this.getExpenseById(expenseId);
+
+    Object.assign(foundExpense, expenseBody);
```
grafit-off, line 79: In this case, user can add any key to our storage, but maybe it's not critical

src/services/users.js
```diff
@@ -0,0 +1,43 @@
...
+    this.data = [];
+  }
+
+  getUsers() {
+    return this.data;
+  }
+
+  createUser(name) {
+    const maxId = Math.max(...this.data.map(user => user.id));
+    const newId = (this.data.length > 0) ? maxId + 1 : 1;
```
grafit-off, line 14: Looks similar like in expenses service, not critical, but I prefer to move that function to utils, or something like that, to avoid duplicates

grafit-off (approved):
GJ ⭐ It's gonna be good to have controllers in my opinion, but maybe it's not critical.

Anna-Tkachenko (approved):
Good job!
Just check @grafit-off comments, and will be good to move corresponding functions for each query to controllers

PR #93, opened 2022-12-08
solution

DarkMistyRoom:


src/services/expenses.js
```diff
@@ -0,0 +1,69 @@
...
+
+  if (from) {
+    foundExpenses = foundExpenses
+      .filter(expense => Date.parse(expense.spentAt) >= Date.parse(from));
+  }
+
+  if (to) {
+    foundExpenses = foundExpenses
+      .filter(expense => Date.parse(expense.spentAt) <= Date.parse(to));
+  }
```
Bohdan-Maksymiuk, line 32: so if you have all params, you will iterate 4 times?
better to do this logic due to 1 iteration only

src/controllers/expenses.js
```diff
@@ -0,0 +1,98 @@
...
+  const newExpense = req.body;
+  const { userId, spentAt, title, amount, category } = newExpense;
+  const foundUser = getUserById(userId);
+
+  if (
+    !title
+    || !spentAt
+    || !amount
+    || !category
+    || !foundUser
```
Bohdan-Maksymiuk, line 43: looks like a candidate for separate variable

Bohdan-Maksymiuk (changes requested):

Virass (approved):

PR #99, opened 2022-12-31
Solution

andreyolin:


src/routes/expenses.js
```diff
@@ -0,0 +1,14 @@
...
+const express = require('express');
+const expensesController = require('../controllers/expenses');
+
+const router = express.Router();
+
+router.get('/', express.json(), expensesController.getAll);
+router.get('/:expenseId', express.json(), expensesController.getOne);
+router.post('/', express.json(), expensesController.create);
+router.delete('/:expenseId', express.json(), expensesController.remove);
+router.patch('/:expenseId', express.json(), expensesController.update);
```
andrii-bassarab, line 12: you can simply simplify this logic, adding Middleware 'express.json()' to app.use -> app.use('/expenses', express.json(), expensesRouter);
```suggestion
router.get('/', expensesController.getAll);
router.get('/:expenseId', expensesController.getOne);
router.post('/', expensesController.create);
router.delete('/:expenseId', expensesController.remove);
router.patch('/:expenseId', expensesController.update);
```
then you should not add it to every request

src/routes/users.js
```diff
@@ -0,0 +1,14 @@
...
+const express = require('express');
+const usersController = require('../controllers/users');
+
+const router = express.Router();
+
+router.get('/', express.json(), usersController.getAll);
+router.get('/:userId', express.json(), usersController.getOne);
+router.post('/', express.json(), usersController.create);
+router.delete('/:userId', express.json(), usersController.remove);
+router.patch('/:userId', express.json(), usersController.update);
```
andrii-bassarab, line 12: same here

src/services/expenses.js
```diff
@@ -0,0 +1,71 @@
...
+  expenses = expenses.filter(expanse => expanse.id !== Number(expenseId));
+}
+
+function updateOne(expenseId, newData) {
+  const expense = getById(expenseId);
+
+  Object.assign(expense, newData);
+
+  return expense;
+}
```
andrii-bassarab, line 62: i think you can simplify this function, passing already found expense in params, and then update this expense, in order not to pass the id and then search expense by this id, because you already found this expense in controller/expense

andrii-bassarab (approved):
GJ!

src/services/expenses.js
andreyolin: good idea! thanks!

denys-danyliuk (approved):

PR #100, opened 2023-01-01
solution

andrii-bassarab:


src/controller/users.js
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const getOne = (req, res) => {
+  const { userId } = req.params;
+
+  if (!userId) {
+    res.sendStatus(400);
+
+    return;
+  }
```
andreyolin, line 18: i think it`s unnecessary if, cause if you have not id yo will not get foundUser

src/controller/users.js
```diff
@@ -0,0 +1,101 @@
...
+const addOne = (req, res) => {
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  if (typeof name !== 'string') {
```
andreyolin, line 40: If i remember correctly you will always have string type from req.body

andreyolin (approved):
gj! to pass some test try to reset users and expences to [] in your createServer.js

denys-danyliuk (approved):

PR #102, opened 2023-01-02
Initial TS commit

vadymbaranov:


tests/user.test.js
```diff
@@ -1,7 +1,5 @@
-'use strict';
-
-const supertest = require('supertest');
-const { createServer } = require('../src/createServer');
+import supertest from 'supertest';
+import { createServer } from '../src/createServer';
```
viktoriia-ovcharenko, line 2: I don't understand why you need it if you use CommonJS

src/services/expenses.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+let expenses = [];
+
+function getAll() {
+  return expenses;
+}
+
+function findExpenseById(expenseId) {
```
viktoriia-ovcharenko, line 9: function should filter expenses by some conditions (for example, return all expenses between dates)

viktoriia-ovcharenko (approved):

tests/user.test.js
vadymbaranov: You're right, that's just a leftover from previous commit where I tried to use imports.

src/services/expenses.js
vadymbaranov: But why? Neither task, nor documentation requires that. This function only checks if the expense with given id, exists or not.

src/controllers/expensesController.js
```diff
@@ -0,0 +1,95 @@
...
+  const { userId, spentAt, title, amount, category, note } = req.body;
+  const foundUser = usersService.findUserById(userId);
+
+  if (!foundUser) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  if (typeof userId !== 'number' || typeof spentAt !== 'string'
```
denys-danyliuk, line 36:
```suggestion
  if (!userId || typeof spentAt !== 'string'
```
do you really need this type of checks or you can go with just an '!'?

src/services/expenses.js
```diff
@@ -0,0 +1,74 @@
...
+
+  if (from) {
+    visibleExpenses = visibleExpenses
+      .filter(expense => expense.spentAt > from);
+  }
+
+  if (to) {
+    visibleExpenses = visibleExpenses
+      .filter(expense => expense.spentAt < to);
+  }
```
denys-danyliuk, line 31: you can combine this two

denys-danyliuk (approved):

PR #103, opened 2023-01-02
v01

Anna-Shapovalova:


taraselnik (approved):
cool! i did it

src/services/expenses.js
```diff
@@ -0,0 +1,80 @@
...
+    return filteredExpenses;
+  }
+
+  if (from && to) {
+    const filteredExpenses = expenses
+      .filter(expense =>
+        expense.spentAt >= from && expense.spentAt <= to,
+      );
+
+    return filteredExpenses;
```
illiarerun, line 23: you can filter by both category AND from or to, don't return it right away

src/services/expenses.js
```diff
@@ -0,0 +1,80 @@
...
+
+function getAll(userId, category, from, to) {
+  if (category) {
+    const filteredExpenses = expenses
+      .filter(expense => expense.category === category);
+
+    return filteredExpenses;
+  }
+
+  if (from && to) {
```
illiarerun, line 17: same here, you can filter either by one of those or them together

PR #112, opened 2023-01-03
solution

Oleksandr-Stryzhavchuk:


taraselnik (approved):
cool

src/services/expenses.js
```diff
@@ -0,0 +1,80 @@
...
+
+function getAll(userId, category, from, to) {
+  if (category) {
+    const filteredExpenses = expenses
+      .filter(expense => expense.category === category);
+
+    return filteredExpenses;
+  }
+
+  if (from && to) {
```
illiarerun, line 17: what if we only pass `from` or only `to`

src/services/expenses.js
```diff
@@ -0,0 +1,80 @@
...
+function init() {
+  expenses = [];
+};
+
+function getAll(userId, category, from, to) {
+  if (category) {
+    const filteredExpenses = expenses
+      .filter(expense => expense.category === category);
+
+    return filteredExpenses;
```
illiarerun, line 14: what if we combine `category` and `from`?

src/services/expenses.js
```diff
@@ -0,0 +1,80 @@
...
+  return expenses;
+};
+
+function getbyId(expenseId) {
+  const foundExpense = expenses.find(expense => expense.id === +expenseId);
+
+  return foundExpense || null;
+}
+
+function create(userId, spentAt, title, amount, category, note) {
```
illiarerun, line 42: pass the whole body as a param, do you really need to describe each field individually?

src/services/expenses.js
```diff
@@ -0,0 +1,80 @@
...
+};
+
+function remove(expenseId) {
+  const filteredExpenses = expenses
+    .filter(expense => expense.id !== +expenseId);
+
+  expenses = filteredExpenses;
+};
+
+function update(expenseId, title) {
```
illiarerun, line 65: what if I want to update not only a title?

src/controllers/expensesController.js
```diff
@@ -0,0 +1,105 @@
...
+    return;
+  }
+
+  expensesService.remove(expenseId);
+  res.sendStatus(204);
+};
+
+function update(req, res) {
+  const { expenseId } = req.params;
+  const { title } = req.body;
```
illiarerun, line 83: again, I can pass not only a title here

src/controllers/expensesController.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  const newExpense = expensesService.create(
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  );
```
illiarerun, line 60: you can just pass the whole body here
```suggestion
  const newExpense = expensesService.create(req.body);
```

illiarerun (changes requested):

PR #113, opened 2023-01-03
initial commit

MikhailKucherina:


src/createServer.js
```diff
@@ -1,11 +1,23 @@
...
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  usersService.runUsers();
+  expensesService.runExpenses();
+
```
denys-danyliuk, line 16: you should reset your arrays, it should fix the issue with failing tests

denys-danyliuk (approved):

PR #114, opened 2023-01-03
Accounting app was added

iryna-lelyk:


src/services/expenses.js
```diff
@@ -0,0 +1,82 @@
...
+  } = expenseDetails;
+
+  const newExpense = {
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+    id: expenses.length + 1,
```
Andriy-Shymkiv, line 55: what if we delete first item from array and then add new one? id will be the same

src/services/users.js
```diff
@@ -0,0 +1,49 @@
...
+function getById(userId) {
+  const foundUser = users.find(user => user.id === +userId);
+
+  return foundUser || null;
+}
+
+function create(name) {
+  const newUser = {
+    name,
+    id: users.length + 1,
```
Andriy-Shymkiv, line 22: same here

Andriy-Shymkiv (approved):
well done 👍

denys-danyliuk (approved):

PR #120, opened 2023-01-03
develop

YaroslavKolbaiev:


src/controllers/expences.js
```diff
@@ -0,0 +1,133 @@
...
+    res.sendStatus(404);
+    return;
+  }
+
+  const selectedExpenses = services.getExpencesByUserId(userId);
+
+  if (!selectedExpenses.length) {
+    res.sendStatus(404);
+    return;
+  }
```
loralevitska, line 40: You can combine it to avoid repeatable code fragment

src/controllers/expences.js
```diff
@@ -0,0 +1,133 @@
...
+
+
+  res.send(newExpense);
+}
+
+export const remove = (req, res) => {
+  const { expenceId } = req.params;
+
+  const foundExpense = services.getExpenceById(expenceId);
+
```
loralevitska, line 84:
```suggestion
```

src/controllers/expences.js
```diff
@@ -0,0 +1,133 @@
...
+    amount,
+    category,
+    note
+  } = req.body;
+
+  if (typeof spentAt !== 'string'
+    || typeof title !== 'string'
+    || typeof amount !== 'number'
+    || typeof category !== 'string'
+    || typeof note !== 'string') {
```
loralevitska, line 117: Move (string types) to array and use [some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) || [every()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)

Avoid such long expressions in `If` block. Use variables and helper functions

src/routes/users.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+import * as usersController from '../controllers/users.js';
+import express from 'express';
+
```
loralevitska, line 5: Remove all extra lines (leave only 1 empty line like divider)

loralevitska (changes requested):
Need some improvements)

loralevitska (approved):

PR #121, opened 2023-01-04
Solution

Ilona-Shulha:


src/controllers/expenses.js
```diff
@@ -0,0 +1,95 @@
...
+
+  if (category) {
+    expenses = expenses.filter(el => el.category === category);
+  }
+
+  if (from && to) {
+    expenses = expenses.filter(el => el.spentAt > from && el.spentAt < to);
+  }
+
+  res.send(expenses);
```
SanyaBratashchuk, line 29: in my opinion controllers shouldn't have as much business logic they should be as simple as it can be. receive a request call a service send a response. So better move it to the services

SanyaBratashchuk: Fix it in all your controllers

SanyaBratashchuk (changes requested):

AlisaVlasova (approved):

PR #122, opened 2023-01-24
solution

Ivan-Nevmerzhytskyi:


src/services/expenses.js
```diff
@@ -0,0 +1,76 @@
...
+
+function getbyId(expenseId) {
+  const foundExpense = expenses.find(expense => expense.id === +expenseId);
+
+  return foundExpense || null;
+}
+
+function create(userId, spentAt, title, amount, category, note) {
+  const newExpense = {
+    id: expenses.length + 1,
```
Bohdan-Maksymiuk, line 43: what if we delete one (not last) and then add?
there will be expenses with duplicated ids

Bohdan-Maksymiuk: try to find max and then increment

Bohdan-Maksymiuk: but don't care, in real life db will handle it instead of you

Bohdan-Maksymiuk (approved):

PR #124, opened 2023-02-04
Solution

shyptia:


src/services/expenses.js
```diff
@@ -0,0 +1,67 @@
...
+      ? expense.spentAt <= to
+      : true;
+
+    return isUserIdMatch && isCategoriesMatch && isFromMatch && isToMatch;
+  });
+};
+
+function addExpense(expenseData) {
+  const newExpense = {
+    id: expenses.length + 1,
```
Oleksandr-Filo, line 33: I think you have to use a separate counter for Id generating, because now if you add 5 expenses to collection, then you remove first two expenses and after removing if you add new expense it will be with Id 4 (not 6). There will two expenses with Id 4 in the collection.

src/services/users.js
```diff
@@ -0,0 +1,49 @@
...
+  users = [];
+}
+
+function getAll() {
+  return users;
+};
+
+function addUser(name) {
+  const newUser = {
+    id: users.length + 1,
```
Oleksandr-Filo, line 15: the same here

Oleksandr-Filo (approved):
looks good, just a few things

src/controllers/expenses.js
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  const expenseData = {
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  };
```
solaryasha, line 47: We should validate the request body. What if I don't send the required fields?
image

solaryasha (changes requested):

src/controllers/expenses.js
```diff
@@ -0,0 +1,127 @@
...
+
+function getAll(req, res) {
+  const { userId, category, from, to } = req.query;
+
+  const searchParams = {
+    userId,
+    category,
+    from,
+    to,
+  };
```
ivan-kadykalo, line 14:
```suggestion
  const searchParams = req.query;
```

src/controllers/expenses.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  const expenseData = {
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  };
```
ivan-kadykalo, line 54:
```suggestion
  const expenseData = req.body;
```

ivan-kadykalo (approved):
Good job 🔥

PR #126, opened 2023-02-06
done!

dmitriyn3679:


src/controllers/expenses.js
```diff
@@ -0,0 +1,81 @@
...
+  res.send(foundExpense);
+};
+
+const add = (req, res) => {
+  const { userId } = req.body;
+
+  const user = usersService.getById(userId);
+
+  if (!user) {
+    res.sendStatus(400);
```
slavanosatskyi, line 32: in this case the required parameter was passed, but the entity was not found. You should rather use 404 code here

slavanosatskyi (changes requested):

slavanosatskyi (approved):
Approved!

PR #129, opened 2023-02-12
Express

Vladyslav317:


src/components/expenses/expenses.service.js
```diff
@@ -0,0 +1,50 @@
+let expenses = [];
+
+export function getAll(userId, category, from, to) {
+  expenses = expenses.filter((expense) => expense.userId === userId);
+  expenses = expenses.filter((expense) => expense.category === category);
+  expenses = expenses.filter((expense) => expense.spentAt >= from);
+  expenses.expenses.filter((expense) => expense.spentAt <= to);
```
marina-tilniak, line 7: you can use one filter here, because now it looks strange and it has bad performance

marina-tilniak (approved):

PR #130, opened 2023-02-13
done

Nazar-Brunarskyi:


src/expenses/controllers.js
```diff
@@ -0,0 +1,95 @@
+'use strict';
+
+const { getUser } = require('../users/services');
+const expensesServices = require('./services');
+
+module.exports.createExpense = (req, res) => {
```
mr-valentos, line 6: It is better to do exports in one place. Like you did with services:
```
module.exports = {
  resetExpenses,
  getExpenses,
  createExpenses,
  getExpense,
  deleteExpense,
  updateExpense,
};

mr-valentos (changes requested):
Almost done. GJ!

PR #135, opened 2023-02-21
Accounting app was implemented

vadiimvooo:


NadiiaKoch (approved):
Good job!

src/controllers/expenses.js
```diff
@@ -0,0 +1,96 @@
...
+const expanseServise = require('../services/expenses');
+const userServise = require('../services/users');
+
+const add = (req, res) => {
+  const { userId, spentAt, title, amount, category, note } = req.body;
+
+  const foundUser = userServise.getById(userId);
+
+  const isValidData
+    = foundUser
```
ivan-kadykalo, line 12: `=` shouldn't start from the new line
```suggestion
  const isValidData = foundUser
```

src/services/expenses.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+let expanses = [];
+let idsExpansesCounter = 0;
+
+function initialExpanses() {
```
ivan-kadykalo, line 6: This name doesn't show what this function does, name it `setDefaultExpanses` or `clearDB` etc

tests/expense.test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
ivan-kadykalo, line 1: We need to disable only max-len in this file

src/services/users.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+let users = [];
+let idsUserCounter = 0;
+
+function initialUsers() {
```
ivan-kadykalo, line 6: the same as for expenses

src/controllers/expenses.js
```diff
@@ -0,0 +1,96 @@
...
+    && typeof spentAt === 'string'
+    && typeof title === 'string'
+    && typeof amount === 'number'
+    && typeof category === 'string'
+    && typeof note === 'string';
```
ivan-kadykalo, line 17: the note is not a necessary field, what will be in the case when req.body will came without this field?

ivan-kadykalo (changes requested):
You are almost there 💪

ivan-kadykalo (approved):
Good job 🔥

PR #138, opened 2023-02-21
solution

Fynkzia:


GabrielMaksimovich (approved):
for me looks good!

src/controllers/expenses.js
```diff
@@ -0,0 +1,123 @@
...
+
+  const foundUser = usersServices.getById(userId);
+
+  const isAllDataProvided = userId
+    && spentAt
+    && title
+    && amount
+    && category;
+
+  const isAllDataValid = foundUser
```
ivan-kadykalo, line 53: `isAllDataValid` made the same as `isAllDataProvided`, cause if some field will be missed, it will not match your check.
For example, if there will be no `title`, it would not match '`string`' type

src/controllers/users.js
```diff
@@ -0,0 +1,103 @@
...
+  const { userId } = req.params;
+  const { name } = req.body;
+
+  if (!userId || !name) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  if (typeof name !== 'string') {
```
ivan-kadykalo, line 73: Combine similar logic
```suggestion
  if (!userId || !name || typeof name !== 'string') {
    res.sendStatus(400);

    return;
  }
```

ivan-kadykalo (approved):
Good job 🔥
Fix those comments in next task

PR #148, opened 2023-02-23
solution

olena-yanovska:


anastasiia-bilska (approved):
Looks good to me :)

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,94 @@
...
+    category,
+  } = req.body;
+
+  const foundUser = userService.findById(userId);
+
+  const isRequiredDataProvided = userId
+    && spentAt
+    && title
+    && amount
+    && category;
```
ivan-kadykalo, line 42: What if the type of data will not match?

ivan-kadykalo (approved):
Good job 🔥

PR #159, opened 2023-03-04
solution

nzrii-kysel:


src/controllers/expenses.js
```diff
@@ -0,0 +1,91 @@
...
+  } catch (err) {
+    res.sendStatus(400);
+  }
+};
+
+const add = (req, res) => {
+  const data = req.body;
+
+  try {
+    if (!usersService.getById(data.userId)) {
```
anna-inozemets, line 34: Maybe it is better to create variable for this

src/controllers/users.js
```diff
@@ -0,0 +1,84 @@
...
+
+    usersService.remove(id);
+    res.sendStatus(204);
+  } catch (err) {
+    res.sendStatus(400);
+  }
+};
+
+const update = (req, res) => {
+  const id = +req.params.id;
```
anna-inozemets, line 60: Maybe all data conversion is better to do in services

src/services/expenses.js
```diff
@@ -0,0 +1,94 @@
...
+
+const validator = require('./expenses.validator');
+
+let expenses = [];
+
+const initiate = (initialValue) => {
+  expenses = initialValue;
+};
+
+const getNewId = () => (
```
anna-inozemets, line 11: It is possible to create one universal function for users and exponses for creating id

anna-inozemets (approved):
Great solution!)

nzrii-kysel:
> Great solution!)

thanks for the review, good remarks :)

Anna-Tkachenko (approved):
Awesome solution, Well Done! ⭐️

PR #163, opened 2023-03-09
solution without MVC

PolinaSukhorukova:


tania-boiar (changes requested):
Well done, for now!
Need to move on and finish the task using MVC and express.Router().
If you want, I could look through it again after that.

src/createServer.js
```diff
@@ -1,11 +1,24 @@
...
-  // Return the server (express app)
+  const app = express();
+
+  app.use(cors());
+
+  userService.getInitialValue();
+  expenseService.getInitialValue();
+
+  app.use('/users', routerUsers);
+  app.use('/expenses', routerExpenses);
```
tania-boiar, line 19: I'd use here middleware `express.json`, like this:
`app.use('/users', express.json(), userRouter);`

src/services/users.js
```diff
@@ -0,0 +1,55 @@
...
+
+let users = [];
+
+const getInitialValue = () => {
+  users = [];
+};
+
+const getAllUsers = () => {
+  return users;
+};
```
tania-boiar, line 13: I'd write this in one line.

src/utils/getFilteredExpenses.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  if (to) {
+    filteredExpenses = expenses.filter(expense => (
+      expense.spentAt <= to
+    ));
+  }
+
+  return filteredExpenses;
+};
```
tania-boiar, line 30: I have `reject` for this(. 
We need to check all `params` in one iteration, not 4.

src/controllers/expenses.js
```diff
@@ -0,0 +1,98 @@
...
+};
+
+module.exports = {
+  expenseController: {
+    getAllExpenses,
+    getOneExpense,
+    addExpense,
+    removeExpense,
+    updateExpense,
+  },
```
tania-boiar, line 97: As I was told, no need to create an object in `module.exports` only where we import it, like this `const expenseController = require('../controllers/expenses');`

tania-boiar (approved):
Well done! ✨ 
You can look through the comments I left for you,
and make some changes if you agree with them.

ArturErmolenko (approved):
Nicely done 🔥

PR #167, opened 2023-03-10
solution

Oleksandr-Maly:


src/controllers/expenses.js
```diff
@@ -0,0 +1,132 @@
...
+    && title
+    && amount
+    && category
+  );
+
+  if (!isAllDataProvided) {
+    res.sendStatus(400);
+
+    return;
+  }
```
OlhaYevstifieieva, line 58: You can also add data validation.
```
typeof userId === 'number'
...and other params...
if (!isAllDataProvided || !foundUser) {
    res.sendStatus(400);
    return;
  }
  ```

File: src/controllers/expenses.js [hunk 2]
```diff
@@ -0,0 +1,132 @@
...
+
+const updateExpense = (req, res) => {
+  const { expenseId } = req.params;
+  const newData = req.body;
+
+  if (!newData) {
+    res.sendStatus(400);
+
+    return;
+  }
```
OlhaYevstifieieva, line 102: Can we combine this with this 👇🏻 into one condition?
```
 if (!expenseId) {
    res.sendStatus(400);

    return;
  }
  ```

File: src/controllers/users.js [hunk 3]
```diff
@@ -0,0 +1,100 @@
...
+const updateUser = (req, res) => {
+  const { userId } = req.params;
+  const { name } = req.body;
+  const invalidData = !userId || !name || (typeof name !== 'string');
+
+  if (invalidData) {
+    res.sendStatus(400);
+
+    return;
+  }
```
OlhaYevstifieieva, line 76: 👍🏻

src/controllers/expenses.js
```diff
@@ -0,0 +1,132 @@
...
+const getOne = (req, res) => {
+  const { expenseId } = req.params;
+
+  if (!expenseId) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const foundExpense = expenseService.getExpenseById(expenseId);
```
Anna-Tkachenko, line 22: should we check the type for `expenseId` value?

src/controllers/users.js
```diff
@@ -0,0 +1,100 @@
...
+
+  usersService.deleteUser(userId);
+  res.sendStatus(204);
+};
+
+const updateUser = (req, res) => {
+  const { userId } = req.params;
+  const { name } = req.body;
+  const invalidData = !userId || !name || (typeof name !== 'string');
```
Anna-Tkachenko, line 70: should we check the type for `userId` value?

Anna-Tkachenko (approved):
Good job! ⭐️

PR #169, opened 2023-03-11
accounting app was done

OlyaZaiets:


src/createServer.js
```diff
@@ -1,11 +1,226 @@
...
+      return;
+    }
+
+    res.send(foundUser);
+  });
+
+  app.post('/users', express.json(), (req, res) => {
+    const { name } = req.body;
+
+    if (typeof name !== 'string') {
```
Coyoote, line 47: Would also be great to check for empty string

ArturErmolenko (approved):
Good job 👍

PR #177, opened 2023-03-12
add accounting app

Anastasiia-Khudych:


imariash86 (approved):
Good job)

src/controllers/expenses.js
```diff
@@ -0,0 +1,120 @@
+'use strict';
+
+const expenseService = require('../services/expenses');
+const userServices = require('../services/users');
+
+const getAll = (req, res) => {
+  const filterParams = req.query;
+
+  const expenses = expenseService.getAll();
+  const filteredExpenses
```
Anna-Tkachenko, line 10: you can also create separate `getFilteredExpenses` method in the service

src/controllers/expenses.js
```diff
@@ -0,0 +1,120 @@
...
+
+  res.send(filteredExpenses);
+};
+
+const getOne = (req, res) => {
+  const { expenseId } = req.params;
+
+  const expense = expenseService.getOne(+expenseId);
+
+  if (isNaN(+expenseId)) {
```
Anna-Tkachenko, line 21: should we check it before trying to get expense?

src/controllers/expenses.js
```diff
@@ -0,0 +1,120 @@
...
+  } = req.body;
+
+  const user = userServices.getOne(+userId);
+
+  if ((typeof spentAt !== 'string')
+    || (typeof title !== 'string')
+    || (typeof amount !== 'number')
+    || (typeof category !== 'string')
+    || (note && typeof note !== 'string')
+    || !user) {
```
Anna-Tkachenko, line 53: if you have several conditions, will be better to move the result of them to separate boolean variable

src/controllers/expenses.js
```diff
@@ -0,0 +1,120 @@
...
+    amount,
+    category,
+    note,
+  } = req.body;
+
+  if ((spentAt && typeof spentAt !== 'string')
+    || (title && typeof title !== 'string')
+    || (amount && typeof amount !== 'number')
+    || (category && typeof category !== 'string')
+    || (note && typeof note !== 'string')) {
```
Anna-Tkachenko, line 88: same here, let's create separate boolean variable that we'll use inside the `if` statement

src/controllers/expenses.js
```diff
@@ -0,0 +1,120 @@
...
+
+  const updatedExpense = expenseService.updateExpense(+expenseId, req.body);
+
+  res.send(updatedExpense);
+};
+
+const deleteExpense = (req, res) => {
+  const { expenseId } = req.params;
+
+  const expense = expenseService.getOne(+expenseId);
```
Anna-Tkachenko, line 102: should we check the type of `expenseId` here?

src/controllers/users.js
```diff
@@ -0,0 +1,89 @@
...
+
+  res.send(users);
+};
+
+const getOne = (req, res) => {
+  const { userId } = req.params;
+
+  const user = userServices.getOne(+userId);
+
+  if (isNaN(+userId)) {
```
Anna-Tkachenko, line 16: will be better to check it before trying to get the user

src/helpers.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const generateId = (arr) => {
```
Anna-Tkachenko, line 3: don't use data type name for variable name

Anna-Tkachenko (approved):

PR #185, opened 2023-03-12
realised accounting-app

andrii-kurash:
Have a great day🔥

### DIFF
--- a/src/services/expense.js
+++ b/src/services/expense.js
@@ -1,0 +1,87 @@
+'use strict';
+
+let expenses = [];
+
+const resetExpenses = () => {
+  expenses = [];
+};
+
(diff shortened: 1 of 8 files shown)


src/services/expense.js
```diff
@@ -0,0 +1,87 @@
...
+};
+
+const getById = (expenseId) => {
+  const foundExpense = expenses.find(({ id }) => id === expenseId);
+
+  return foundExpense || null;
+};
+
+const addExpense = (expenseData) => {
+  const maxId = Math.max(...expenses.map(expense => expense.id), 0);
```
Anna-Tkachenko, line 52: will be better to create separate util function for getting max id

Anna-Tkachenko (approved):

PR #193, opened 2023-03-13
solution

victorionov:


src/controllers/expenses.js
```diff
@@ -0,0 +1,87 @@
...
+
+  const filteredExpenses = expenseService.getAll(queryParams);
+
+  res.send(filteredExpenses);
+};
+
+const getOne = (req, res) => {
+  const { expenseId } = req.params;
+
+  const currentExpense = expenseService.getByExpenseId(expenseId);
```
Anna-Tkachenko, line 17: should we check correct type for `expenseId`?

src/controllers/users.js
```diff
@@ -0,0 +1,82 @@
...
+    return;
+  }
+
+  res.send(currentUser);
+};
+
+const add = (req, res) => {
+  const { name } = req.body;
+
+  if (!name || typeof name !== 'string') {
```
Anna-Tkachenko, line 28: if you have several checks will be better to move the result to separate boolean variable

src/services/expenses.js
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const getByExpenseId = (expenseId) => {
+  const currentExpense = expenses.find(expence => expence.id === expenseId);
+
+  return currentExpense || null;
+};
+
+const create = (body) => {
+  const newId = Math.max(...expenses.map(({ id }) => id), 0) + 1;
```
Anna-Tkachenko, line 33: will be better to move the logic for getting new id to separate function

src/createServer.js
```diff
@@ -1,11 +1,25 @@
 'use strict';
 
-// const express = require('express');
+const express = require('express');
+const cors = require('cors');
+
+const { router: usersRouter } = require('./routes/users');
+const { router: expensesRouter } = require('./routes/expenses');
```
Anna-Tkachenko, line 7: check file for routes, they're empty

Anna-Tkachenko (changes requested):

PR #198, opened 2023-04-02
solution

vitalii-babiienko:


src/controllers/expenses.js
```diff
@@ -0,0 +1,108 @@
...
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(foundExpense);
+};
+
+const add = (req, res) => {
```
SerhiiKirik, line 51: I think, it would be better name
```suggestion
const create = (req, res) => {
```

SerhiiKirik (approved):
LGTM 🔥

PR #200, opened 2023-04-05
server functionality done

eugenekh81:
please review)


nazarmatsevych (changes requested):
test

src/createServer.js
```diff
@@ -1,11 +1,251 @@
...
+        : true;
+
+      return isUserIdMatch && isCategoriesMatch && isFromMatch && isToMatch;
+    });
+
+    res.send(filteredExpenses);
+  });
+
+  app.get('/expenses/:id', (req, res) => {
+    const { id } = req.params;
```
DarkMistyRoom, line 46: check if `id` was passed

src/createServer.js
```diff
@@ -1,11 +1,251 @@
...
+    const {
+      userId,
+      spentAt,
+      title,
+      amount,
+      category,
+      note,
+    } = req.body;
+
+    const requiredFields = ['userId', 'spentAt', 'title', 'amount', 'category'];
```
DarkMistyRoom, line 69: move it to a util file

src/createServer.js
```diff
@@ -1,11 +1,251 @@
...
+      return;
+    }
+
+    const foundUser = users.find(user => user.id === +userId);
+
+    if (!foundUser) {
+      res.sendStatus(400);
+
+      return;
+    }
```
DarkMistyRoom, line 88: why do you do it? it is an `add `method.

src/createServer.js
```diff
@@ -1,11 +1,251 @@
...
+      && isNaN(Date.parse(spentAt))
+      && typeof title !== 'string'
+      && typeof amount !== 'number'
+      && typeof category !== 'string'
+      && typeof note !== 'string'
+    ) {
+      res.sendStatus(400);
+
+      return;
+    }
```
DarkMistyRoom, line 101: combine it with `isReqFieldsValid`

src/createServer.js
```diff
@@ -1,11 +1,251 @@
...
+    };
+
+    expenses.push(newExpense);
+
+    res.status(201);
+    res.send(newExpense);
+  });
+
+  app.patch('/expenses/:id', express.json(), (req, res) => {
+    const { id } = req.params;
```
DarkMistyRoom, line 115: check if `id` was passed

DarkMistyRoom (changes requested):
Separate the code on routes, controllers, and services like in the video. Now it is hard to read

src/createServer.js
eugenekh81: these is a requirement in the tests, if the user is not found in the create new expense request, it shoud return status code 400
Screenshot 2023-04-17 at 02 27 41

src/services/expenses.js
```diff
@@ -0,0 +1,65 @@
...
+
+    const categoriesMatch = categories
+      ? categories.includes(expense.category)
+      : true;
+
+    const datesMatch = (from && to)
+      ? expense.spentAt >= from && expense.spentAt <= to
+      : true;
+
+    return userIdMatch && categoriesMatch && datesMatch;
```
DarkMistyRoom, line 25: ok, but we can use `if` to return false on the first `false` result and avoid next calculations

PR #201, opened 2023-04-07
solution

kirill-vypirovskyi:


stanislavcodes (approved):
🔥

src/createServer.js
```diff
@@ -1,11 +1,21 @@
...
+const express = require('express');
+const { router: usersRouter } = require('./routes/users');
+const { router: expenseRouter } = require('./routes/expenses');
+const { initialize: inizializeUsers } = require('./services/users');
+const { initialize: inizializeExpenses } = require('./services/expenses');
+
+const app = express();
+
+app.use('/users', express.json(), usersRouter);
+app.use('/expenses', express.json(), expenseRouter);
```
And678, line 12: Instead of adding `express.json()` to every route, add it above using `app.use()`
```suggestion
app.use(express.json());

app.use('/users', usersRouter);
app.use('/expenses', expenseRouter);
```

src/models/users.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const { generateId } = require('../functions/generateId');
+
+class Expenses {
```
And678, line 5: Forgot to change the name here.

```suggestion
class Users {
```

src/controllers/users.js
```diff
@@ -0,0 +1,77 @@
...
+
+  if (!foundUser) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  if (typeof name !== 'string') {
+    res.sendStatus(422);
+  }
```
And678, line 64: Just like before, add `return;` after `res.sendStatus` to prevent further execution of the function.

package.json
```diff
@@ -25,6 +25,7 @@
     "supertest": "^6.3.0"
   },
   "dependencies": {
-    "express": "^4.18.2"
+    "express": "^4.18.2",
+    "uuid": "^9.0.0"
```
And678, line 29: Make sure to avoid adding unneeded deps. `uuid` is not needed, as requirement calls for integer ids.
```suggestion
```

src/routes/expenses.js
```diff
@@ -0,0 +1,23 @@
...
+
+const hasParams = (req, _res, next) => {
+  if (Object.keys(req.query).length > 0) {
+    next();
+  } else {
+    next('route');
+  }
+};
+
+router.get('/', hasParams, expenseController.getByParameters);
```
And678, line 16: Good work figuring out a custom solution using next('route') :1st_place_medal:. Unfortunately, introducing custom logic to your routing is error-prone and can get real complicated in big projects.

Move this logic to the controller (use only one controller method). This is okay, as `.getAll` is only 2 lines long.

And678 (changes requested):
Really nice!

Only minor issues present, keep up the great work :fire:

src/controllers/expenses.js
```diff
@@ -0,0 +1,170 @@
...
+    || !isDateValid(spentAt)
+    || typeof title !== 'string'
+    || typeof amount !== 'number'
+    || typeof category !== 'string'
+    || typeof note !== 'string'
+  ) {
+    res.sendStatus(400);
+
+    return;
+  }
```
And678, line 87: Nice work validating requests.
The issues is, request parameters should be checked before you query your database (or in-memory model). Move this check before `userService.getById(userId);`.

Also this applies to all of your endpoints, make sure to move validation before your queries/searches/deletes/etc.

src/functions/generateId.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const generateId = () => {
+  return Math.trunc(Math.random() * 100000);
+};
```
OleksiiBlezniuk, line 5: Nice, but it is possible to get existed id using this logic =)

OleksiiBlezniuk (approved):

PR #203, opened 2023-04-10
Solution

stanislavcodes:


rashyd-hasratov (approved):
Good job! Check my comments please, maybe it will make your solution match the requirements better.

Oleksandr-kopaevich (approved):
Hey @stanislavcodes what an amazing job done here. I love your code splitting, structure, and organization, can't suggest much to improve. All tests passed, so approved, but keep in mind for the future: the first thing we do in every request before calling business logic is validating request parameters to be sure that they meet our expectations.

PR #206, opened 2023-04-27
Solution

lilia-obushenko:


KuharchukDmytro (approved):
gj!

src/services/expenses.js
```diff
@@ -0,0 +1,81 @@
...
+
+  if (categories) {
+    visibleExpenses = expenses.filter(
+      ({ category }) => categories.includes(category)
+    );
+  }
+
+  if (from) {
+    visibleExpenses = expenses.filter(({ spentAt }) => spentAt > from);
+  }
```
OleksiiBlezniuk, line 33: Why can't I combine filter options? Let's instead of filtering initial array filter already filtered one

src/services/users.js
```diff
@@ -0,0 +1,51 @@
...
+    name,
+  };
+
+  users.push(newUser);
+
+  return newUser;
+};
+
+const removeUser = (userId) => {
+  users = users.filter(user => user.id !== +userId);
```
OleksiiBlezniuk, line 33: I believe that it's better to pass converted to number value to this method

OleksiiBlezniuk (changes requested):

src/services/expenses.js
```diff
@@ -0,0 +1,87 @@
...
+  return expenses;
+};
+
+const getAllExpenses = (
+  id,
+  categories,
+  from,
+  to,
+) => {
+  let visibleExpenses = expenses;
```
OleksiiBlezniuk, line 19: But we can't see it =)
I guess `filteredExpenses` will be more suitable here

OleksiiBlezniuk (approved):

PR #207, opened 2023-04-30
solution

Zelinskyi-Serhii:


OleksiiBlezniuk:
> i just forgot to add package.json file Tests don't work corectly. They need that i have empty array after each test.

Screenshot 2023-05-04 at 17 41 42

src/createServer.js
```diff
@@ -1,13 +1,25 @@
...
+
+  app.use('/users', express.json(), usersRouter);
+  app.use('/expenses', express.json(), expensesRouter);
+
+  return app;
 }
 
+// createServer().listen(3000, () => {
+//   global.console.log('Server are running');
+// });
```
OleksiiBlezniuk, line 21: remove commented code

src/services/expenses.js
```diff
@@ -0,0 +1,108 @@
...
+
+let countExpensesId = 1;
+
+const getExpenses = ({ from, to, category, userId }) => {
+  const startFrom = Date.parse(from);
+  const end = Date.parse(to);
+  let filteredExpenses = expenses;
+
+  if (startFrom) {
+    const result = filteredExpenses.filter(expense => {
```
OleksiiBlezniuk, line 13: I think there is no need to create variable for result, you can write it to `filteredEpxenses`

src/services/expenses.js
```diff
@@ -0,0 +1,108 @@
...
+  return foundExpense || null;
+};
+
+const createExpense = ({
+  userId,
+  spentAt,
+  title,
+  amount,
+  category,
+  note = 'Any notes',
```
OleksiiBlezniuk, line 67: Don't set initial note here, let it be null
What if we'll handle this case on FE part and show a different message?

src/controllers/users.js
```diff
@@ -0,0 +1,96 @@
...
+const userService = require('../services/users.js');
+
+const getAll = (req, res) => {
+  const users = userService.getAll();
+
+  res.send(users);
+};
+
+const getOne = (req, res) => {
+  const userId = +req.params.userId;
```
OleksiiBlezniuk, line 12:
```suggestion
  const userId = Number(req.params.userId);
```

src/controllers/users.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  const newUser = userService.addUser(name);
+
+  res.status(201);
+  res.send(newUser);
+};
+
+const removeUser = (req, res) => {
+  const userId = +req.params.userId;
```
OleksiiBlezniuk, line 47:
```suggestion
  const userId = Number(req.params.userId);
```

OleksiiBlezniuk (changes requested):

PR #208, opened 2023-05-02
accounting app v2 done

rashyd-hasratov:


src/controllers/expenses.js
```diff
@@ -0,0 +1,133 @@
...
+
+  const foundUser = userService.getById(userId);
+
+  if (!foundUser) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newExpense = expenseService.create(req.body);
```
And678, line 74: Better to pass only the required fields here instead of entire, possibly invalid body.

And678: Applies to the rest of code, too.

src/createServer.js
```diff
@@ -1,11 +1,24 @@
...
-  // Add a routes to the server
-  // Return the server (express app)
+  userService.init();
+  expenseService.init();
+
+  const app = express();
+
+  app.use(cors());
+  app.use('/users', express.json(), userRouter);
+  app.use('/expenses', express.json(), expenseRouter);
```
And678, line 19: Remember about DRY. You can `use` `express.json()` only once before your routes.
```suggestion
  const app = express();
  
  app.use(express.json());
  app.use(cors());
  app.use('/users', userRouter);
  app.use('/expenses', expenseRouter);
```

src/services/expenses.js
```diff
@@ -0,0 +1,87 @@
+'use strict';
+
+class Expenses {
+  constructor() {
+    this.data = [];
+    this.idCounter = 0;
+  }
+
+  init() {
```
And678, line 9: Having and `init` method together with identical constructor is redundant. You need to reset your data for tests, though, so you can call it reset. (same for user service)
```suggestion
  reset() {
```

And678 (changes requested):
Good job, but looks like you forgot to push `package.json`. That's why the tests [here](https://github.com/mate-academy/node_accounting-app-v2/pull/208/checks) fail. Also fails on my computer too.

Please push `package.json` with the `cors` package included.
Feel free to reach out in QnA if you have any questions.

src/controllers/expenses.js
```diff
@@ -0,0 +1,153 @@
...
+    return;
+  }
+
+  expenseService.update(foundExpense, {
+    spentAt: spentAt || foundExpense.spentAt,
+    title: title || foundExpense.title,
+    amount: amount || foundExpense.amount,
+    category: category || foundExpense.category,
+    note: note || foundExpense.note,
+  });
```
OleksiiBlezniuk, line 139: but you're using Object.assign in update, just passed fields to update, no need to pass same properties here

src/controllers/users.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
+
+const { service: userService } = require('../services/users');
+
+const getAll = (req, res) => {
+  const foundUsers = userService.getAll();
+
+  res.statusCode = 200;
```
OleksiiBlezniuk, line 8: default statusCode

src/services/expenses.js
```diff
@@ -0,0 +1,87 @@
...
+      categories,
+      from,
+      to,
+    } = filterOptions;
+
+    let filteredExpenses = this.data;
+
+    if (userId) {
+      filteredExpenses = filteredExpenses.filter(expense => (
+        expense.userId === +userId
```
OleksiiBlezniuk, line 26: Can we convert it outside this method?

src/services/expenses.js
```diff
@@ -0,0 +1,87 @@
...
+  removeById(expenseId) {
+    this.data = this.data.filter(({ id }) => id !== +expenseId);
+  }
+
+  update(expense, partsToUpdate) {
+    Object.assign(expense, partsToUpdate);
+  }
+}
+
+module.exports.service = new Expenses();
```
OleksiiBlezniuk, line 87: But why do we export not a class but it's instance?

OleksiiBlezniuk (approved):

PR #209, opened 2023-05-03
accounting-app was implement

KuharchukDmytro:


src/controllers/users.js
```diff
@@ -0,0 +1,88 @@
+'use strict';
+
+const { UsersModel } = require('../models/users.js');
```
sergey-lipikhin, line 3:
```suggestion
const { usersModel } = require('../models/users.js');
```

It should be named with a lowercase first letter since it is an object, not class. And maybe would be more descriptive to rename `usersModel` to `userService`

src/models/users.js
```diff
@@ -0,0 +1,59 @@
...
+  getAll() {
+    return this.users;
+  };
+
+  getById(userId) {
+    return this.users.find(user => user.id === userId) || null;
+  }
+
+  addUser(name) {
+    const maxId = this.users.lengt
```
sergey-lipikhin, line 21:
```suggestion
    const maxId = this.users.length
```

So it always gets the the same for all objects. But the tests have passed :)

sergey-lipikhin (approved):

src/controllers/expenses.js
```diff
@@ -0,0 +1,102 @@
...
+  const isUserExist = userService.getById(Number(userId));
+
+  if (!title || !isUserExist) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newExpense = expensesService.create(
+    userId,
```
juljaVysotska, line 46: Use destructuring assignment for the arguments passed to the create function to make the code more readable and less prone to errors.

src/services/usersService.js
```diff
@@ -0,0 +1,59 @@
...
+  getAll() {
+    return this.users;
+  };
+
+  getById(userId) {
+    return this.users.find(user => user.id === userId) || null;
+  }
+
+  addUser(name) {
+    const maxId = this.users.length
```
juljaVysotska, line 21: Use `const maxId = this.users.reduce((max, user) => Math.max(max, user.id), -1);` instead of using `Math.max` with spread operator over `map` to simplify and optimize the code.

src/controllers/expenses.js
```diff
@@ -0,0 +1,102 @@
+'use strict';
+
+const { expensesService } = require('../services/expensesService.js');
+const { userService } = require('../services/usersService.js');
+
+const getAll = (req, res) => {
+  const filteredExpenses = expensesService.getAll(req.query);
+
+  res.statusCode = 200;
```
OleksiiBlezniuk, line 9: this is default status code

src/controllers/expenses.js
```diff
@@ -0,0 +1,102 @@
...
+  const {
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
+
+  const isUserExist = userService.getById(Number(userId));
```
OleksiiBlezniuk, line 37:
```suggestion
  const user = userService.getById(Number(userId));
```

src/services/expensesService.js
```diff
@@ -0,0 +1,127 @@
...
+      categories,
+      from,
+      to,
+    } = requestBody;
+
+    let filteredExpenses = this.expenses;
+
+    if (userId) {
+      filteredExpenses = filteredExpenses.filter(
+        expense => expense.userId === +userId,
```
OleksiiBlezniuk, line 24:
```suggestion
        expense => expense.userId === Number(userId),
```

src/services/expensesService.js
```diff
@@ -0,0 +1,127 @@
...
+    this.expenses = [];
+  }
+
+  getAll(requestBody) {
+    const {
+      userId,
+      categories,
+      from,
+      to,
+    } = requestBody;
```
juljaVysotska, line 18: U can make destructuring at once in function parameters.

src/services/expensesService.js
```diff
@@ -0,0 +1,127 @@
...
+      category: category || foundExpense.category,
+      note: note || foundExpense.note,
+    });
+
+    return foundExpense;
+  }
+}
+
+module.exports = {
+  expensesService: new ExpensesModel(),
```
OleksiiBlezniuk, line 126: I don't think that it's a good idea to export instance of this class here

src/services/usersService.js
```diff
@@ -0,0 +1,59 @@
...
+      return false;
+    }
+
+    this.users = this.users.filter(user => user.id !== userId);
+
+    return true;
+  }
+
+  updateUser(userId, name) {
+    const foundUser = this.users.find(user => user.id === userId) || null;
```
OleksiiBlezniuk, line 45: use existing method

src/services/usersService.js
```diff
@@ -0,0 +1,59 @@
...
+      name,
+    };
+
+    this.users.push(newUser);
+
+    return newUser;
+  }
+
+  removeUser(userId) {
+    if (!this.users.some(user => user.id === userId)) {
```
OleksiiBlezniuk, line 35: You have similar method

OleksiiBlezniuk (changes requested):

juljaVysotska (changes requested):

OleksiiBlezniuk (approved):

juljaVysotska (approved):

PR #210, opened 2023-05-03
Solution

timoxa2t:


src/routers/expenses.js
```diff
@@ -0,0 +1,127 @@
...
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
+
+  if (!(
+    userId !== undefined
+    && userService.getUserById(userId.toString())
```
vladyslav-mosor, line 35: Better to create const for that

src/routers/expenses.js
```diff
@@ -0,0 +1,127 @@
...
+    && spentAt !== undefined
+    && title !== undefined
+    && amount !== undefined
+    && category !== undefined
+  )) {
+    res.sendStatus(400);
+
+    return;
+  }
```
vladyslav-mosor, line 44: I would rather just implement checking with ! and || operators
```suggestion
   if (!userId || !user || !title || !amount || !category) {
    res.sendStatus(400);

    return;
  }
```

vladyslav-mosor (changes requested):
GJ overall but I found some issues to fix. Please, move the functions from router to separate controllers so your code will be more structured according to MVC + pay attention to additional comments.

src/routers/expenses.js
timoxa2t: that was my first thought, but values 0 and '' also could be there

src/services/expenses.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
+
+let nextId = 1;
+let expences = [];
+
+function getExpenses(userId, categories, from, to) {
+  let res = [...expences];
```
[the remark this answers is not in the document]
timoxa2t: it depends on the params, so I couldn't assign either of this

vladyslav-mosor (approved):

src/services/user.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+// const uuid = require('uuid');
+
+let nextId = 1;
```
And678, line 5: Nice work making sure that ids don't repeat :fire:

src/createServer.js
```diff
@@ -1,11 +1,21 @@
...
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  resetUsers();
+  resetExpenses();
+
```
And678, line 14: You can move `express.json()` call here from your routers
```suggestion
app.use(express.json());
```

src/controllers/expenses.js
```diff
@@ -0,0 +1,132 @@
...
+    'note',
+  ];
+
+  const changes = {};
+
+  for (const key of allowedKeys) {
+    if (req.body[key] !== undefined) {
+      changes[key] = req.body[key];
+    }
+  }
```
And678, line 113: Nice! No unneeded keys in your update :fire: 
Also, you could move this logic to a separate function to reuse somewhere else (not needed on this instance, but in general)

And678 (approved):
Awesome, hard to find anything to complain about :heavy_check_mark:

PR #214, opened 2023-05-04
Accounting app

gretoq:


src/controllers/expenses.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  if (to) {
+    expenses = expenses.filter(expense => expense.spentAt <= to);
+  }
+
+  if (categories) {
+    expenses
+      = expenses.filter(expense => categories.includes(expense.category));
+  }
```
And678, line 30: This logic belongs to service, not the controller. Move all filtering/search logic there.

In a real life scenario, `.getAll` would do some database call inside. Downloading the entire dataset and then filtering in your controller code would be inefficient, so you should just pass your filtering options to service and do your filtering there.

src/services/users.js
```diff
@@ -0,0 +1,47 @@
...
+const getById = (userId) => {
+  const foundUser = users.find(({ id }) => id === +userId);
+
+  return foundUser;
+};
+
+const create = (userBody) => {
+  const newUser = {
+    id: getNewId(users),
+    ...userBody,
```
And678, line 20: What if `userBody` contains some unneeded fields? Because of this, user can push any gibberish into your data. Make sure to filter your input body, and only save the data required by documentation.

And678 (changes requested):
Good job, nice idea having new id function in utils. Fix the two issues below and this is done :100:

And678 (approved):
Nice work :heavy_check_mark:

PR #215, opened 2023-05-04
solution

valeriiayarzhemska:


Ambulence (approved):
Great!

src/services/expensesService.js
```diff
@@ -0,0 +1,89 @@
...
+const updateExpense = ({
+  expenseId,
+  requestBody,
+}) => {
+  let updatedExpense = getExpenseById(expenseId);
+
+  updatedExpense = {
+    ...updatedExpense,
+    ...requestBody,
+  };
```
And678, line 77: This doesn't update the expense in your `expenses` array, you can use `Object.assign` for that

package.json
```diff
@@ -25,6 +25,8 @@
     "supertest": "^6.3.0"
   },
   "dependencies": {
-    "express": "^4.18.2"
+    "cors": "^2.8.5",
+    "express": "^4.18.2",
+    "express-router": "0.0.1"
```
And678, line 30: Express already contains the router, and this package is not used in your code anyway. Including unmaintained libraries is a security hazard, avoid it.
```suggestion
```

src/controllers/expensesController.js
```diff
@@ -0,0 +1,90 @@
...
+  const { expenseId } = req.params;
+  const foundExpense = expensesService.getExpenseById(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const requestBody = req.body;
```
And678, line 74: By this logic, user can add some non standard properties here, make sure to filter out properties that are not defined in docs.

src/services/expensesService.js
```diff
@@ -0,0 +1,89 @@
...
+    categories,
+    from,
+    to,
+  } = query;
+
+  let filteredExpenses = expenses;
+
+  if (userId) {
+    filteredExpenses = expenses
+      .filter(expense => expense.userId === Number(userId));
```
And678, line 25: Move this and the rest of `Number(...)` conversions to controller level. Dealing with input data is the area of concern for the controller, not the service.

src/createServer.js
```diff
@@ -1,11 +1,24 @@
...
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  usersService.resetUsers();
+  expensesService.resetExpenses();
+
+  app.use(cors());
+  app.use('/users', express.json(), usersRouter);
+  app.use('/expenses', express.json(), expensesRouter);
```
And678, line 19: Can be simplified here.
```suggestion

  app.use(cors());
  app.use(express.json());
  app.use('/users', usersRouter);
  app.use('/expenses', expensesRouter);
```

And678 (changes requested):
Overall, good job, but some things need to be fixed below

PR #216, opened 2023-05-04
Accounting app

Anton-Tretiak:


src/services/usersService.js
```diff
@@ -0,0 +1,49 @@
...
+
+const getUserById = (userId) => {
+  const foundUser = users.find(user => user.id === +userId);
+
+  return foundUser || null;
+};
+
+const createUser = (name) => {
+  const newUser = {
+    id: Math.max(...users.map(user => user.id), 0) + 1,
```
And678, line 21: Good stuff, but remember about DRY. This `Math.max(...data.map(...` logic is the same for both users and expenses. Move it to function and place it in separate `utils` folder.

src/createServer.js
```diff
@@ -1,11 +1,24 @@
...
-  // Add a routes to the server
-  // Return the server (express app)
+  usersService.getInitial();
+  expensesService.getInitial();
+
+  const app = express();
+
+  app.use(cors());
+  app.use('/users', express.json(), usersRouter);
+  app.use('/expenses', express.json(), expensesRouter);
```
And678, line 19: A little simpler this way :)

```suggestion

  app.use(cors());
  app.use(express.json());
  app.use('/users', usersRouter);
  app.use('/expenses', expensesRouter);
```

src/services/expensesService.js
```diff
@@ -0,0 +1,79 @@
...
+  expenses = expenses.filter(({ id }) => id !== +expenseId);
+};
+
+const updateExpense = (expenseId, body) => {
+  let foundExpense = expenses.find(({ id }) => id === +expenseId) || null;
+
+  foundExpense = {
+    ...foundExpense,
+    ...body,
+  };
```
And678, line 67: This returns the correct response, but doesn't actually update the actual expense. Try getting the expense after updating, you'll see that it didn't update.

To fix, you can use [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)

And678 (changes requested):
Good work, one important bug and a couple of issues to fix below.

And678 (changes requested):
Good work, one important bug and a couple of issues to fix below.

PR #219, opened 2023-05-06
accounting-app

Ambulence:


src/controllers/expenses.js
```diff
@@ -0,0 +1,77 @@
...
+const getAll = (req, res) => {
+  const expenses = expensesService.getAll(req.query);
+
+  res.send(expenses);
+};
+
+const getOne = (req, res) => {
+  const { expenseId } = req.params;
+
+  const foundExpense = expensesService.getById(+expenseId);
```
juljaVysotska, line 15:
```suggestion
  const foundExpense = expensesService.getById(Number(expenseId));
```
For better reading and understanding your code

src/services/users.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+const remove = (userId) => {
+  users = users.filter(user => user.id !== userId);
+};
+
+const update = (userId, name) => {
+  const foundUser = getById(userId);
+
+  Object.assign(foundUser, { name });
```
juljaVysotska, line 37: Same situation. maybe we can return object with new user data from this method, not just mutate?

juljaVysotska (changes requested):
Good job! I found some issues, please check and fix them. Best regards :)

PR #220, opened 2023-05-06
Solution

mstrilec:


src/controllers/expensesController.js
```diff
@@ -0,0 +1,129 @@
+'use strict';
+
+const userService = require('../services/users');
```
juljaVysotska, line 3: Use destructuring to import multiple exports from the same module: const { getUser } = require('../services/users');

juljaVysotska (changes requested):
Good job! I like your code! I found some moments, please fix them. Best regards 🧡

PR #222, opened 2023-05-06
Solution

sasha-dranuta:


And678 (changes requested):
Hello, app doesn't start, make sure it runs correctly and tests pass
image

PR #223, opened 2023-05-07
AccoutingApp task done

MarkBabich:


src/servises/expenses.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function update(expenseId, requestBody) {
+  const expense = getById(expenseId);
+  const {
+    spentAt = expense.spentAt,
+    title = expense.title,
+    amount = expense.amount,
+    category = expense.category,
+    note = expense.note,
```
And678, line 79: Nice, no unneeded properties from req.body :fire:

And678 (changes requested):
Good job, but some things to improve below :arrow_down:

PR #224, opened 2023-05-07
solution

NikitaStambul:


And678 (changes requested):
Hello! Make sure that the [tests](https://github.com/mate-academy/node_accounting-app-v2/pull/227/checks) are passing.
Currently, they are failing because you forgot to push package.json after installing new packages ("cors" and "uuid"). Do that, and there will also be some more tests to fix after that.

PR #227, opened 2023-05-08
Develop

yaroslav-vasilyev:


src/createServer.js
```diff
@@ -1,11 +1,30 @@
...
   // Use express to create a server
   // Add a routes to the server
   // Return the server (express app)
+
+  const app = express();
+
+  app.use(cors());
+
+  app.use('/users', express.json(), usersRouter);
+  app.use('/expenses', express.json(), expensesRouter);
```
And678, line 20:
```suggestion
  app.use(express.json());

  app.use('/users', usersRouter);
  app.use('/expenses', expensesRouter);
```

And678 (changes requested):
Hello! Make sure that the [tests](https://github.com/mate-academy/node_accounting-app-v2/pull/227/checks) are passing. 
Currently, they are failing because you forgot to push `package.json` after installing a new package ("cors"). Do that, and there will also be some more tests to fix after that.

PR #230, opened 2023-05-10
node accounting - solution

Yevhenii0536:


src/services/users.js
```diff
@@ -0,0 +1,47 @@
...
+
+function getById(id) {
+  const foundUser = users.find(user => user.id === id);
+
+  return foundUser || null;
+}
+
+function create(name) {
+  const newUser = {
+    id: Math.floor(Math.random() * 100),
```
And678, line 21: This will cause duplicate ids sooner or later. Instead, you can make a variable at the top of this file, near `let users = [];`, that keeps track of the latest id. After that you can just do this:
```suggestion
    id: latestId++,
```

Real databases have similar functionality used for simple integer ids (for example in [postgres](https://chartio.com/resources/tutorials/how-to-define-an-auto-increment-primary-key-in-postgresql/))

src/controllers/expenses.js
```diff
@@ -0,0 +1,129 @@
...
+    return;
+  }
+
+  const {
+    userId = foundExpense.userId,
+    spentAt = foundExpense.spentAt,
+    title = foundExpense.title,
+    amount = foundExpense.amount,
+    category = foundExpense.category,
+    note = foundExpense.note,
```
And678, line 101: Nice! :fire:

And678 (changes requested):
Good job, a couple of things to improve below, and this is done :100:

PR #231, opened 2023-05-13
Solution

MykhailoIvchenko:
There are some tests which were not passed, but after many checks I came to a conclusion, that server doesn't stop between tests so the next test operates with data created during previous test.
Also I've changed in tests the type of id from number to string to use uuid library.


src/services/expenses.js
```diff
@@ -0,0 +1,69 @@
...
+  return expenses.filter(expense => expense.userId === userId);
+}
+
+function getAllForCategory(expencesList, category) {
+  return expencesList.filter(expense => expense.category === category);
+}
+
+function getAllBetweenDates(expencesList, from, to) {
+  return expencesList.filter(expense => {
+    const timestamp = new Date(expense.spentAt).getTime();
```
nazarmatsevych, line 50: Use a single return statement instead of if-else. Simplify the return value of the filter callback.

tests/user.test.js
```diff
@@ -27,7 +27,7 @@ describe('User', () => {
       expect(response.body)
         .toEqual(
           expect.objectContaining({
-            id: expect.any(Number),
+            id: expect.any(String),
```
nazarmatsevych, line 30: The 'id' field is expected to be a String, but the task description does not specify this change. Make sure the change is aligned with the task requirements.

tests/expense.test.js
```diff
@@ -38,7 +38,7 @@ describe('Expense', () => {
       expect(response.body)
         .toEqual(
           expect.objectContaining({
-            id: expect.any(Number),
+            id: expect.any(String),
```
nazarmatsevych, line 41: It seems like the ID type has been changed from Number to String. Make sure this change was intentional and is consistent across the codebase.

src/services/users.js
```diff
@@ -0,0 +1,45 @@
...
+function getAll() {
+  return users;
+}
+
+function getById(userId) {
+  const foundUser = users.find(user => user.id === userId);
+
+  return foundUser || null;
+}
+
```
nazarmatsevych, line 17: Missing validation for the 'name' parameter. It should check if the provided name is not empty or undefined before creating a new user.

src/services/users.js
```diff
@@ -0,0 +1,45 @@
...
+  // count++;
+
+  return newUser;
+}
+
+function remove(userId) {
+  users = users.filter(user => user.id !== userId);
+}
+
+function update({ id, name }) {
```
nazarmatsevych, line 35: Missing validation for the 'id' and 'name' parameters. It should check if the provided id and name are not empty or undefined before updating the user.

src/controllers/users.js
```diff
@@ -0,0 +1,97 @@
...
+    res.sendStatus(422);
+
+    return;
+  }
+
+  update({
+    id, name,
+  });
+
+  res.send(foundUser);
```
nazarmatsevych, line 92: You should send the updated user object instead of the old one to the client. Call getById(id) again after updating the user to get the updated user object.

src/controllers/users.js
```diff
@@ -0,0 +1,97 @@
...
+
+  const foundUser = getById(id);
+
+  if (!foundUser) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const { name } = req.body;
```
nazarmatsevych, line 80: You should check if the name is not empty or only contains whitespace, as this would create an invalid user.

src/controllers/expenses.js
```diff
@@ -0,0 +1,112 @@
...
+
+const getAllExpenses = (req, res) => {
+  let expenses = getAll();
+  const { userId } = req.query;
+  const { categories } = req.query;
+  const { from, to } = req.query;
+
+  switch (true) {
+    case !!userId:
+      if (userId) {
```
nazarmatsevych, line 25: Remove unnecessary 'if (userId)' check, as it is already checked in the switch case.

src/controllers/expenses.js
```diff
@@ -0,0 +1,112 @@
...
+        expenses = getAllForUser(userId);
+      }
+
+    case !!categories:
+      if (categories) {
```
nazarmatsevych, line 30: Remove unnecessary 'if (categories)' check, as it is already checked in the switch case.

src/controllers/expenses.js
```diff
@@ -0,0 +1,112 @@
...
+        expenses = getAllForCategory(expenses, categories);
+      }
+
+    case (!!from && !!to):
+      if (from && to) {
```
nazarmatsevych, line 35: Remove unnecessary 'if (from && to)' check, as it is already checked in the switch case.

src/controllers/expenses.js
```diff
@@ -0,0 +1,112 @@
...
+
+const { getById: getUserById } = require('../services/users');
+
+const getAllExpenses = (req, res) => {
+  let expenses = getAll();
+  const { userId } = req.query;
+  const { categories } = req.query;
+  const { from, to } = req.query;
+
+  switch (true) {
```
nazarmatsevych, line 23: Use if statements instead of switch cases without 'break;' to avoid unintended behavior.

src/controllers/expenses.js
```diff
@@ -0,0 +1,112 @@
...
+const createExpense = (req, res) => {
+  const user = getUserById(req.body.userId);
+
+  if (!user) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  if (Object.keys(req.body) === 0) {
```
nazarmatsevych, line 68: Use 'Object.keys(req.body).length === 0' to check if the req.body is empty.

nazarmatsevych (changes requested):
image

Create a function for example

`function init() {
  expenses = [];
}`

and call it inside createServer

`function createServer() {
  const app = express();

  app.use(cors());

  userServices.init();
  expenseServices.init();

  app.use('/users', express.json(), usersRouter);
  app.use('/expenses', express.json(), expensesRouter);

  return app;
}`


if you have any questions - feel free to ask them in fe_chat

src/services/expenses.js
MykhailoIvchenko: Done

tests/user.test.js
MykhailoIvchenko: I've changed it for purpose to use uuid library. In documentation id is integer, and if it is necessary I'd create variable count to use it as id. But I don't think, this is a good enough solution.

tests/expense.test.js
MykhailoIvchenko: It was intentional and in the app ids are strings.

src/services/users.js
MykhailoIvchenko: Seeems like I've deleted it with console statements)) Done

src/services/users.js
MykhailoIvchenko: Function updateUser calls getById which return null if id is invalid or user wasn't found. Name check was added now

src/controllers/expenses.js
MykhailoIvchenko: Done

src/controllers/expenses.js
MykhailoIvchenko: Done

src/controllers/expenses.js
MykhailoIvchenko: Done

src/controllers/expenses.js
MykhailoIvchenko: Done

src/controllers/expenses.js
MykhailoIvchenko: Done

nazarmatsevych (approved):
Well done!

PR #234, opened 2023-06-01
created a node accounting app

prokivskaaa:


src/controllers/usersController.js
```diff
@@ -0,0 +1,86 @@
...
+const getAllUsers = (req, res) => {
+  const users = usersModel.getAllUsers();
+
+  return res.status(200).send(users);
+};
+
+const getUserById = (req, res) => {
+  const { userId } = req.params;
+
+  const user = usersModel.getUserById(Number(userId));
```
SerhiiKirik, line 14: It's better to check if userId is a valid number (not NaN) after converting it using Number().

PR #237, opened 2023-06-02
app to track expenses added

DenysKrysanov:


src/helpers/helpers.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+const getNewId = (data) => {
+  return Math.max(...data.map((item) => item.id), 0) + 1;
```
SerhiiKirik, line 4: Change the starting value from 0 to - in Math.max() to avoid returning 0 as the first new id.

src/controllers/usersController.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  res.send(foundUser);
+};
+
+const create = (req, res) => {
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
```
SerhiiKirik, line 35: Check for the type of 'name' to make sure it's a string before checking if it's empty.

src/controllers/usersController.js
```diff
@@ -0,0 +1,91 @@
...
+
+  res.sendStatus(204);
+};
+
+const update = (req, res) => {
+  const userId = Number(req.params.userId);
+  const foundUser = userService.getUser(userId);
+  const { name } = req.body;
+
+  if (!name || isNaN(userId) || typeof name !== 'string') {
```
SerhiiKirik, line 66: Split the conditional statement into separate checks for better error handling and readability.

src/controllers/usersController.js
```diff
@@ -0,0 +1,91 @@
...
+    res.sendStatus(404);
+
+    return;
+  }
+
+  userService.update({
+    id: userId,
+    name,
+  });
+  res.send(foundUser);
```
SerhiiKirik, line 82: After updating the user, send the updated user object in the response instead of the old one.

src/controllers/expensesController.js
```diff
@@ -0,0 +1,129 @@
...
+  if (!foundUser) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newExpense = expensesService
+    .create(userId, spentAt, title, amount, category, note);
+
+  res.statusCode = 201;
```
SerhiiKirik, line 59: you can use method for it
```suggestion
  res.status(201);
```

src/controllers/usersController.js
SerhiiKirik, line 82:
```suggestion
  const newUser = userService.update({
    id: userId,
    name,
  });
  
  res.send(newUser);
```

src/createServer.js
```diff
@@ -1,11 +1,22 @@
...
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  clearExpenses();
+  clearUsers();
+
+  app.use('/users', express.json(), userRouter);
+  app.use('/expenses', express.json(), expensesRouter);
```
SerhiiKirik, line 17:
```suggestion
  app.use(express.json());
  app.use('/users', userRouter);
  app.use('/expenses', expensesRouter);
```

src/services/expenses.js
```diff
@@ -0,0 +1,90 @@
...
+
+function getAll(params) {
+  const {
+    userId,
+    categories,
+    from,
+    to,
+  } = params;
+
+  let filteredExpenses = expenses;
```
SerhiiKirik, line 19: why you need to create new variable if you can change just expenses array

SerhiiKirik (changes requested):
Great 🔥 
Let's make a few improvements 🧑‍💻

src/controllers/expensesController.js
```diff
@@ -0,0 +1,129 @@
...
+    id: expenseId,
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  });
+
+  res.send(foundExpense);
```
SerhiiKirik, line 120:
```suggestion
  const upratedExpense = expensesService.update({
    id: expenseId,
    userId,
    spentAt,
    title,
    amount,
    category,
    note,
  });

  res.send(upratedExpense);
```

src/services/expenses.js
```diff
@@ -0,0 +1,90 @@
...
+}
+
+function remove(id) {
+  expenses = expenses.filter(expense => expense.id !== id);
+}
+
+function update({ id, ...data }) {
+  const expense = getById(id);
+
+  Object.assign(expense, data);
```
SerhiiKirik, line 79: update shold return new `expense`

SerhiiKirik (approved):
Great 🔥

PR #238, opened 2023-06-02
solved

PashaMalyshkin:


src/controllers/users.js
```diff
@@ -0,0 +1,75 @@
...
+
+  const foundUser = usersServices.getUserById(userId);
+
+  if (typeof name !== 'string') {
+    res.sendStatus(422);
+
+    return;
+  }
+
+  Object.assign(foundUser, { name });
```
SerhiiKirik, line 64: the same

SerhiiKirik (changes requested):
Nice 🔥 

src/controllers/users.js
```diff
@@ -0,0 +1,76 @@
...
+
+  if (typeof name !== 'string') {
+    res.sendStatus(422);
+
+    return;
+  }
+
+  usersServices.updateUser(foundUser, { name });
+
+  res.send(foundUser);
```
SerhiiKirik, line 67: After update you should return updatedUser

SerhiiKirik (approved):
LGTM 🔥

PR #240, opened 2023-06-02
completed with one test failure, later i will fix it

Zad0rnik:


src/services/expenseServices.js
```diff
@@ -0,0 +1,76 @@
...
+
+function getExpenseById(id) {
+  const foundExpense = expenses.find((expense) => expense.id === +id);
+
+  return foundExpense;
+};
+
+function createExpense(userId, spentAt, title, amount, category, note) {
+  const newExpense = {
+    id: expenses.length + 1,
```
liliia-livandovskaya, line 45: Use a more robust method for generating IDs, like UUID, to avoid potential collisions when creating new expenses.
Or make a separate function for find max id & plus 1 to this value

src/controllers/userController.js
```diff
@@ -0,0 +1,82 @@
+'use strict';
+
+const userServices = require('../services/userServices');
+
+function getUsers(req, res) {
```
liliia-livandovskaya, line 5: consider using arrow function

liliia-livandovskaya (approved):
Great job! Pay attention to the comments

PR #243, opened 2023-06-03
task solved

M1k1ta:


src/routes/expenses.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
```
SerhiiKirik, line 1: Consider removing the 'use strict' directive, as it's not necessary in ES6 modules.

src/routes/users.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
```
SerhiiKirik, line 1: Since you are using ES6 features such as const, you can remove the 'use strict' directive as it is not needed in ES6 modules.

src/controllers/users.js
```diff
@@ -0,0 +1,94 @@
...
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newUser = {
+    id: Math.random(),
```
SerhiiKirik, line 36: Consider using a different method to generate unique ids, such as a counter or a library like uuid.

src/controllers/users.js
```diff
@@ -0,0 +1,94 @@
+'use strict';
+
+let users = [];
+
+const getAllUsers = (req, res) => {
+  if (res) {
```
SerhiiKirik, line 6:
```suggestion
  if (req) {
```

src/controllers/users.js
```diff
@@ -0,0 +1,94 @@
...
+    res.send(users);
+  }
+
+  return users;
+};
+
+const getOne = (req, res) => {
+  const { userId } = req.params;
+  const foundUser = users.find(user => user.id === +userId);
```
SerhiiKirik, line 15: In all cases
```suggestion
  const foundUser = users.find(user => user.id === Number(userId));
```

src/controllers/users.js
SerhiiKirik: Or make a function like `getNextId` where you find max id & plus 1

src/createServer.js
```diff
@@ -1,11 +1,23 @@
...
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  app.use(cors());
+  app.use('/users', express.json(), usersRouter);
+  app.use('/expenses', express.json(), expensesRouter);
```
SerhiiKirik, line 15:
```suggestion

  app.use(cors());
  app.use(express.json());
  app.use('/users', usersRouter);
  app.use('/expenses', expensesRouter);
```

src/controllers/expenses.js
```diff
@@ -0,0 +1,118 @@
...
+    expenses = expenses.filter(({ category }) => (
+      categories.includes(category)
+    ));
+  }
+
+  if (from && to) {
+    expenses = expenses.filter(expense => {
+      const expenseDate = new Date(expense.spentAt);
+      const fromDate = new Date(from);
+      const toDate = new Date(to);
```
SerhiiKirik, line 24: You should parse the dates before filtering. Parsing dates inside the filter function is inefficient.
You create it each iteration

SerhiiKirik (changes requested):
Well done 🔥 
Also, split with services

src/controllers/users.js
SerhiiKirik: It's not main, but your code should be clean & with best practice
Make a function for it, try to handle it!

src/controllers/expenses.js
```diff
@@ -0,0 +1,122 @@
+'use strict';
+
+const { getUsers } = require('./users');
+const { getNextId } = require('../getNextId');
+
```
SerhiiKirik, line 5: Create a service for logic & move logic witch related to the `service`
Split your logic here & in the `users.js` file

SerhiiKirik (changes requested):
Great 🔥 
Let's improve your solution 🧑‍💻

PR #245, opened 2023-06-04
Solution

yuliia-anisimova:


src/routes/expenses.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
```
SerhiiKirik, line 1: Consider removing 'use strict' since it's not necessary in ES6 modules.

PR #249, opened 2023-06-04
solution

StasLukianchuk:


SerhiiKirik (changes requested):
Nice 🔥 
Split it with MVC architectural pattern 🧑‍💻

src/routes/expenses.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
```
SerhiiKirik, line 1: Consider removing the 'use strict' directive as it is unnecessary in ECMAScript modules.
But don't change it here

src/controllers/users.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  const newUser = userService.create(name);
+
+  res.statusCode = 201;
+  res.send(newUser);
+}
+
+function update(req, res) {
+  const { userId } = req.params;
```
SerhiiKirik, line 40: Check if the `name` parameter is passed in the request body. If not, send a 400 status code.

src/controllers/users.js
```diff
@@ -0,0 +1,74 @@
...
+  const { name } = req.body;
+  const foundUser = userService.getById(userId);
+
+  Object.assign(foundUser, { name });
```
SerhiiKirik, line 44: Before updating the user object, check if the user exists. If not, send a 404 status code.

src/controllers/expenses.js
```diff
@@ -0,0 +1,83 @@
...
+
+  if (!title || !hasUser) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newExpense = expenseService.create(req.body);
+
+  res.statusCode = 201;
```
SerhiiKirik, line 41: Use `res.status(201)` instead of `res.statusCode = 201` to set the status code.

src/controllers/expenses.js
```diff
@@ -0,0 +1,83 @@
...
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  Object.assign(foundExpense, data);
+
+  res.statusCode = 200;
```
SerhiiKirik, line 58: Use `res.status(200)` instead of `res.statusCode = 200` to set the status code.
```suggestion
  res.status(200);
```

src/controllers/users.js
SerhiiKirik: also, make it with the `update` method

src/controllers/users.js
```diff
@@ -0,0 +1,74 @@
...
+
+  userService.update({
+    id: userId, name,
+  });
+
+  res.send(foundUser);
```
SerhiiKirik, line 50:
```suggestion
  const updatedUser = userService.update({
    id: userId, name,
  });

  res.send(updatedUser);
```

src/routes/expenses.js
```diff
@@ -0,0 +1,14 @@
...
+
+const express = require('express');
+const expenseController = require('../controllers/expenses.js');
+
+const router = express.Router();
+
+router.get('/', expenseController.getFiltered);
+router.get('/:expenseId', expenseController.getById);
+router.post('/', express.json(), expenseController.create);
+router.patch('/:expenseId', express.json(), expenseController.update);
```
SerhiiKirik, line 11: you already handle it in the `createServer`
```suggestion
router.post('/', expenseController.create);
router.patch('/:expenseId', expenseController.update);
```

src/services/expenses.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+let expenses = [];
+
+function getFiltered(data) {
+  const { userId, categories, from, to } = data;
+
+  if (userId) {
+    expenses = expenses.filter(expense => +userId === expense.userId);
```
SerhiiKirik, line 9: use `Number` instead `+`/
in all cases
```suggestion
    expenses = expenses.filter(expense => Number(userId) === expense.userId);
```

src/services/expenses.js
```diff
@@ -0,0 +1,67 @@
...
+
+function getById(expenseId) {
+  const foundExpense = expenses.find((expense) => expense.id === +expenseId);
+
+  return foundExpense || null;
+}
+
+function create(data) {
+  const newExpense = {
+    id: Math.random(expenses.length + 1),
```
SerhiiKirik, line 33: why `Math.random`?
Use `uuid` lib or make a separate function where you find max id & make a plus one to this value
Reuse this!

src/services/users.js
```diff
@@ -0,0 +1,49 @@
...
+
+function getById(userId) {
+  const foundUser = users.find((user) => user.id === +userId);
+
+  return foundUser || null;
+}
+
+function create(name) {
+  const newUser = {
+    id: Math.random(users.length + 1),
```
SerhiiKirik, line 17: the same

SerhiiKirik (changes requested):
Nice 🔥 

src/controllers/users.js
```diff
@@ -0,0 +1,86 @@
...
+    return;
+  }
+
+  if (!foundUser) {
+    res.statusCode(404);
+
+    return;
+  }
+
+  Object.assign(foundUser, { name });
```
SerhiiKirik, line 56: why do you need this if you make update in 58line?

SerhiiKirik (approved):
Well done 🔥

PR #250, opened 2023-06-04
solution

Olekk17:


src/controllers/users.js
```diff
@@ -0,0 +1,85 @@
...
+
+  if (!name) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newUser = userService.create(name);
+
+  res.statusCode = 201;
```
SerhiiKirik, line 35: Use `res.status(201)` instead of `res.statusCode = 201` for better readability and consistency.

src/routes/expenses.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
```
SerhiiKirik, line 1: Remove 'use strict'; as it's not necessary in ES6 modules.

src/routes/users.js
```diff
@@ -0,0 +1,17 @@
...
+
+router.get('/users', userController.getAll);
+
+router.get('/users/:userId', userController.getOne);
+
+router.post('/users', express.json(), userController.add);
+
+router.delete('/users/:userId', userController.remove);
+
+router.patch('/users/:userId', express.json(), userController.update);
```
SerhiiKirik, line 15:
```suggestion
router.post('/users', userController.add);

router.delete('/users/:userId', userController.remove);

router.patch('/users/:userId', userController.update);
```

src/createServer.js
```diff
@@ -1,11 +1,23 @@
...
+
+const userService = require('./services/users');
+const expenseService = require('./services/expenses');
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
```
SerhiiKirik, line 13: you can just make a 
```suggestion
  app.use(express.json());
```

src/routes/expenses.js
Olekk17: Linter was arguing for that, I did it only because of that

PR #252, opened 2023-06-05
solution

Sbavdik:


src/routes/expense.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
```
SerhiiKirik, line 1: Consider removing 'use strict' since it's not necessary in ES6 modules.

src/services/expense.js
```diff
@@ -0,0 +1,87 @@
...
+
+  return newExpense;
+};
+
+const removeExpense = (expenseId) => {
+  expenses = expenses.filter(({ id }) => id !== expenseId);
+};
+
+const updateExpense = (expenseId, expenseNewData) => {
+  let expense = getById(expenseId);
```
SerhiiKirik, line 70: The updateExpense function is not updating the expense in the expenses array. You should find the index of the expense, then update it directly in the expenses array.

src/controllers/expense.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
+
+const expenseService = require('../services/expense');
+const userServise = require('../services/user');
```
SerhiiKirik, line 4:
```suggestion
const userService = require('../services/user');
```

src/controllers/expense.js
```diff
@@ -0,0 +1,101 @@
...
+
+const getFiltered = (req, res) => {
+  const {
+    userId,
+    categories,
+    from,
+    to,
+  } = req.query;
+
+  const expenses = expenseService.getFiltered(+userId, categories, from, to);
```
SerhiiKirik, line 14: In all cases
```suggestion
  const expenses = expenseService.getFiltered(Number(userId), categories, from, to);
```

src/controllers/expense.js
```diff
@@ -0,0 +1,101 @@
...
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(foundExpense);
+};
+
+const addExpense = (req, res) => {
```
SerhiiKirik, line 39: Unnecessary to add some names to the `add` or `delete`
```suggestion
const add = (req, res) => {
```

src/services/expense.js
```diff
@@ -0,0 +1,87 @@
...
+  if (category) {
+    filteredExpenses = filteredExpenses.filter(expense => (
+      category.includes(expense.category)));
+  }
+
+  if (from) {
+    const fromDate = new Date(from);
+
+    filteredExpenses = filteredExpenses.filter(expense => {
+      const expenseDate = new Date(expense.spentAt);
```
SerhiiKirik, line 26: DRY
Try to find some another solution

SerhiiKirik (changes requested):
Well done 🔥 

src/controllers/expense.js
```diff
@@ -0,0 +1,103 @@
...
+const getFiltered = (req, res) => {
+  const {
+    userId,
+    categories,
+    from,
+    to,
+  } = req.query;
+
+  const expenses = expenseService
+    .getFiltered(Number(userId), categories, from, to);
```
SerhiiKirik, line 15:
```suggestion
  const expenses = expenseService.getFiltered(
    Number(userId), categories, from, to
  );
```

SerhiiKirik (approved):
Great 🔥

PR #253, opened 2023-06-05
Solution

Kavon-Skela:


src/controllers/expenseController.js
```diff
@@ -0,0 +1,110 @@
...
+
+function getAllExpenses(req, res) {
+  const {
+    userId,
+    categories,
+    from,
+    to,
+  } = req.query;
+
+  const filteredExpenses = expenseServices
```
SerhiiKirik, line 14: Check if `userId` is provided before calling `getExpenses` function to avoid unnecessary function calls and to return a 400 response if `userId` is not provided.

src/controllers/expenseController.js
```diff
@@ -0,0 +1,110 @@
...
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.statusCode = 200;
+  res.send(foundExpense);
+}
+
+function createExpense(req, res) {
```
SerhiiKirik, line 41: you can just make it short
```suggestion
function create(req, res) {
```

src/controllers/expenseController.js
```diff
@@ -0,0 +1,110 @@
...
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
+
+  const foundUser = userServices.getByUserId(userId);
+
+  if (!foundUser) {
+    res.sendStatus(400);
```
SerhiiKirik, line 54: Return status code 404 instead of 400 if foundUser is not found, as the requirement states 'Return 404 if expected entity doesn't exist'.
```suggestion
    res.sendStatus(404);
```

src/controllers/userController.js
```diff
@@ -0,0 +1,91 @@
...
+
+  const user = userServices.getByUserId(userId);
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.statusCode = 200;
```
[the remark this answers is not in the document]
SerhiiKirik: do it in all cases

src/routes/expenseRouter.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
```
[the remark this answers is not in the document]
SerhiiKirik: But don't change it here

SerhiiKirik (changes requested):
Nice 🔥 

src/routes/expenseRouter.js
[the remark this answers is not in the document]
Kavon-Skela: image
linter doesn't allow me to remove the 'use strict' instruction

src/controllers/expenseController.js
Kavon-Skela: Maybe Im doing smthng wrong, but tests fail in that way
image

SerhiiKirik: Oh, my but, it's ok

src/services/expenseServices.js
```diff
@@ -0,0 +1,101 @@
...
+      .filter(expense => (
+        expense.spentAt > from && expense.spentAt < to
+      ));
+  }
+
+  if (from && !to) {
+    filteredExpenses = filteredExpenses
+      .filter(expense => (
+        expense.spentAt >= from
+      ));
```
SerhiiKirik, line 35:
```suggestion
      .filter(expense => expense.spentAt >= from);
```

SerhiiKirik (approved):
LGTM 🔥

PR #256, opened 2023-06-05
almost done, without one test

iHateAntichrist:


SerhiiKirik (changes requested):
Good job 🔥
Split it with MVC architectural pattern 🧑‍💻

PR #260, opened 2023-06-06
accounting app

roshirin:


src/routes/users.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
```
SerhiiKirik, line 1: The 'use strict' directive is not necessary in this case, as it is automatically enabled in ES6 modules.
But don't change it here

src/createServer.js
```diff
@@ -1,11 +1,24 @@
...
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  app.use(cors());
+  app.use(express.json());
+  app.use('/users', usersRouter);
+  app.use('/expenses', expensesRouter);
+
+  clearUsers();
```
SerhiiKirik, line 18: Consider moving 'clearUsers()' and 'clearExpenses()' to a separate function, such as 'initializeData()', and call it from within 'createServer()'. This way, you separate the responsibility of data initialization from server creation.

src/controllers/expenses.js
```diff
@@ -0,0 +1,85 @@
...
+
+  if (!isValidUser || !spentAt || !title) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const expense = expenseServices.createExpense(req.body);
+
+  res.statusCode = 201;
```
SerhiiKirik, line 44: you can use the method for it
```suggestion
  res.status(201);
```

src/services/expenses.js
```diff
@@ -0,0 +1,84 @@
...
+let expenses = [];
+
+const filterExpenses = (filters) => {
+  const { userId, categories, from, to } = filters;
+
+  let tempExpenses = [...expenses];
+
+  if (userId) {
+    tempExpenses = tempExpenses.filter(expense => (
+      String(expense.userId) === userId
```
SerhiiKirik, line 14: would be better to compare Number
```suggestion
      expense.userId === Number(userId)
```

SerhiiKirik (approved):
Nice 🔥

PR #265, opened 2023-06-06
solution

Goto-Jesus:


src/routers/users.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const express = require('express');
+const usersController = require('../controllers/users.js');
+
+const usersRouter = express.Router();
+
+usersRouter.get('/', usersController.getAll);
+usersRouter.get('/:userId', usersController.getOne);
```
liliia-livandovskaya, line 9: Consider using a more descriptive name for the route parameter, such as ':id' instead of ':userId'. This will make the code more readable and maintainable.

liliia-livandovskaya (changes requested):
Great job! 👍 
Pay attention to the comments

PR #275, opened 2023-07-07
add task solution

maksym-nemera:
When I added routes, some tests failed. In one file all tests have been done.
But it works correctly, I've tested it on Postman

### DIFF
--- a/src/routes/expenses.js
+++ b/src/routes/expenses.js
@@ -1,0 +1,128 @@
+'use strict';
+
+const express = require('express');
+const router = express.Router();
+const { users } = require('./users.js');
+
+let expenses = [];
+
+router.get('/', (req, res) => {
+  const {
+    userId,
+    from,
+    to,
+    categories,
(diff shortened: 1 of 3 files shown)


src/routes/expenses.js
```diff
@@ -0,0 +1,128 @@
...
+    });
+  }
+
+  if (userId && categories) {
+    filteredExpenses = expenses.filter(expense => (
+      categories.includes(expense.category)
+    ));
+  }
+
+  res.send(filteredExpenses);
```
roman-mirzoian, line 41: Here and in other places, the array is returned to a straight line, although the JSON object should be returned.

roman-mirzoian (approved):
I checked the operation of the application through Postman - I agree, the results are returned correctly, but not completely :) 
Do not forget that you should return JSON objects, and not just an array/data object. 
Also, don't forget to install headers if the task requires it.

PR #276, opened 2023-07-18
done

generatedman:


33Orange (changes requested):
Please fix all linter errors before push 
image

src/createServer.js
```diff
@@ -1,12 +1,19 @@
...
+const { router: expenseRouter } = require('./routes/expenses.js');
 
-function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
-}
+const createServer = () => {
+  const app = express();
+
```
33Orange, line 9: You can create 2 functions that will reset your data arrays.
`setInitialUsers()` should change the users array to an empty array;
`setInitialExpenses()` should change the expenses array to an empty array;

Save this function in their services and call in the `createServer()` function.

It will help you pass the tests

src/routes/users.js
```diff
@@ -0,0 +1,82 @@
...
+  res.send(users);
+});
+
+router.get('/:id', (req, res) => {
+  const { id } = req.params;
+
+  const foundUser = users.find(user => user.id === +id);
+
+  if (!foundUser) {
+    res.sendStatus(404)
```
33Orange, line 18: You forgot about the error message.
```
```

src/routes/users.js
```diff
@@ -0,0 +1,82 @@
...
+  res.send(foundUser);
+});
+
+router.delete('/:id', (req, res) => {
+  const { id } = req.params;
+
+  const foundUser = users.find(user => user.id === +id);
+
+  if (!foundUser) {
+    res.sendStatus(404)
```
33Orange, line 32: Same here and in all other cases

src/routes/users.js
```diff
@@ -0,0 +1,82 @@
...
+  const { id } = req.params;
+
+  const foundUser = users.find(user => user.id === +id);
+
+  if (!name) {
+    res.sendStatus(400);
+
+    return;
+  }
+
```
33Orange, line 74: 1. Check if the name exists in the request before you try to find the user. 
2. Where is the error when the user isn't found?
image

```suggestion
  const { name } = req.body;
  const { id } = req.params;

  if (!name) {
    res.sendStatus(400);

    return;
  }
 
  const foundUser = users.find(user => user.id === +id);

```

src/routes/expenses.js
```diff
@@ -0,0 +1,129 @@
...
+    spentAt,
+    category,
+    note
+  } = req.body;
+
+  const foundUser = users.find(user => user.id === +userId);
+
+  if (!foundUser) {
+    res.status(400);
+    res.send({});
```
33Orange, line 58: Same about the msg

src/routes/expenses.js
```diff
@@ -0,0 +1,129 @@
...
+});
+
+router.get('/:id', (req, res) => {
+  const { id } = req.params;
+
+  const foundExpense = expenses.find(expense => expense.id === +id);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
```
33Orange, line 85: Same

33Orange (changes requested):
You still have the linter errors.

image

PR #278, opened 2023-07-20
made task node_accounting-app-v2

vladyslav-lapshyn:
I can't pass some tests in IDE, but I check it on [postman](https://www.postman.com/), and it works

### DIFF
--- a/src/controllers/expenses.js
+++ b/src/controllers/expenses.js
@@ -1,0 +1,99 @@
+'use strict';
+
+const expensesService = require('../services/expenses.js');
+const userService = require('../services/users.js');
+
+const getAll = (req, res) => {
+  const {
+    userId,
+    from,
+    to,
+    categories,
+  } = req.query;
+
+  const expenses = expensesService.getAll({
+    userId,
+    from,
+    to,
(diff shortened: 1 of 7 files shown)


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,99 @@
...
+
+  res.sendStatus(204);
+};
+
+const update = (req, res) => {
+  const { expenseId } = req.params;
+  const foundExpense = expensesService.getById(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
```
naziamoff, line 85: would be better to make errors more informative

src/services/expenses.service.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
+
+let expenses = [];
+
+function getAll({
```
naziamoff, line 5: getAllByFilters

naziamoff (approved):
looks good 👍 

One important thing - your want your errors to be maximally informative. This way it will be easier to understand what went wrong if it did -and debug it faster

PR #279, opened 2023-07-21
Develop

okosohor:


src/expenses.js
```diff
@@ -0,0 +1,107 @@
...
+});
+
+router.get('/:expenseId', (req, res) => {
+  const { expenseId } = req.params;
+  const foundExpense = expenses.find(expense => expense.id === +expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
```
naziamoff, line 66: your errors have to be more informative. You want them to be maximally clear, as this makes your code much easier to debug and maintain

naziamoff (changes requested):
not bad, but let's now create separate directories for controllers, services and routes, and move our files accordingly.

Bad project architecture is probably the last thing you would like to see on the real project. It makes your code terribly difficult to support

PR #283, opened 2023-07-24
solution

Yuliya1907:


src/services/users.js
```diff
@@ -0,0 +1,52 @@
...
+
+const getAll = () => {
+  return users;
+};
+
+const getById = (userId) => {
+  const foundUser = users.find(user => user.id === userId);
+
+  return foundUser || null;
+};
```
naziamoff, line 17: let's add some error handling and throw an error if user not found

naziamoff (approved):
Looks good 👍 
Few issues found

PR #287, opened 2023-07-24
initial commit

Shevtsooov:


33Orange (changes requested):
Let's create separate directories for controllers, services, and routes, and move our files accordingly to improve readability. 

For example :

`routers/users.js`

`app.get('/users', UserController.getUsers);`

`controllers/users.js`

```
getUsers(req, res) => {
  // logic
}
```

PR #288, opened 2023-07-25
add solution

IrynaBalandiukh:


src/controllers/user.controller.js
```diff
@@ -0,0 +1,95 @@
+'use strict';
+
+const { userService } = require('../services/user.service');
+
+const getAll = (_, res) => {
```
33Orange, line 5:
```suggestion
const getAll = (req, res) => {
```

src/controllers/user.controller.js
```diff
@@ -0,0 +1,95 @@
...
+
+  res.send(users);
+};
+
+const getById = (req, res) => {
+  const { userId } = req.params;
+
+  if (!userId) {
+    res.sendStatus(400);
+
```
33Orange, line 16: Let's add some message 

```
return 400 with any message if required parameter is not passed
```

src/controllers/user.controller.js
```diff
@@ -0,0 +1,95 @@
...
+    return;
+  }
+
+  const foundUser = userService.getById(Number(userId));
+
+  if (!foundUser) {
+    res.sendStatus(404);
+
```
33Orange, line 24: Same

src/controllers/user.controller.js
```diff
@@ -0,0 +1,95 @@
...
+  res.send(foundUser);
+};
+
+const add = (req, res) => {
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
+
```
33Orange, line 36: Same here and in all other cases

33Orange (changes requested):
Almost done

src/createServer.js
```diff
@@ -1,13 +1,20 @@
...
+const express = require('express');
+const { userRouter } = require('./routers/user.router');
+const { expenseRouter } = require('./routers/expense.router');
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
```
33Orange, line 9: You can create 2 functions that will reset your data arrays.
`setInitialUsers()` should change the users array to an empty array;
`setInitialExpenses()` should change the expenses array to an empty array;

Save this function in their services and call in the `createServer()` function.

It will help you pass the tests

PR #289, opened 2023-07-25
solution

annholovko:


naziamoff (changes requested):
Looks not bad, but let's now create separate directories for controllers, services, and routes, and move our files and methods accordingly.

Bad project architecture is probably the last thing you would like to see on a real project. It makes your code terribly difficult to support

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,125 @@
...
+    || typeof amount !== 'number'
+    || typeof category !== 'string'
+    || typeof note !== 'string'
+    || Object.keys(req.body).length > 5
+    || Object.keys(req.body).length === 0
+  ) {
+    res.status(422).send('Incorrect Data Format');
+
+    return;
+  }
```
33Orange, line 114: I think the problem in this check

image

33Orange (changes requested):
Almost done

naziamoff (approved):

PR #290, opened 2023-07-26
Develop

niaYaro:


src/controllers/expenseController.js
```diff
@@ -0,0 +1,82 @@
+'use strict';
+
+const { ExpenseService } = require('../services/expensesService');
+const { UserService } = require('../services/usersService');
+
+const expenseService = new ExpenseService();
+const userService = new UserService();
```
33Orange, line 7: If you wanna use classes they should be [singletons](https://www.freecodecamp.org/news/singleton-design-pattern-with-javascript/) in this case

src/controllers/expenseController.js
```diff
@@ -0,0 +1,82 @@
...
+const getAllExpensesController = (req, res) => {
```
33Orange, line 8: Is it necessary to add "Controller" in every function name?

```suggestion
const getAllExpenses = (req, res) => {
```

src/controllers/userController.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+const { UserService } = require('../services/usersService');
+const userService = new UserService();
+const getAllUsersController = (req, res) => {
```
33Orange, line 5: Same here and in all other cases

```suggestion
const getAllUsers = (req, res) => {
```

src/controllers/userController.js
```diff
@@ -0,0 +1,84 @@
...
+  const users = userService.getAll();
+
+  res.statusCode = 200;
```
33Orange, line 8:
```suggestion
  res.sendStatus(200);
```

src/controllers/userController.js
```diff
@@ -0,0 +1,84 @@
...
+  const { id } = req.params;
+  const foundUser = userService.getOne(id);
+
+  if (!foundUser) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.statusCode = 200;
```
33Orange, line 22: Same here and in all other cases

src/controllers/userController.js
```diff
@@ -0,0 +1,84 @@
...
+    res.sendStatus(404);
+
+    return;
+  }
+
+  if (!name) {
+    res.sendStatus(422);
+
+    return;
+  }
```
33Orange, line 71: First of all, check the data exist

```suggestion
  const { id } = req.params;
  const { name } = req.body;
  
   if (!name) {
    res.sendStatus(422);

    return;
  }
  
  const userShouldUpdate = userService.getOne(id);

  if (!userShouldUpdate) {
    res.sendStatus(404);

    return;
  }

 
```

src/controllers/expenseController.js
```diff
@@ -0,0 +1,82 @@
...
+  res.status(200);
+  res.send(expenses);
+};
+
+const getOneExpenseController = (req, res) => {
+  const { expenseId } = req.params;
+  const foundExpense = expenseService.getOne(+expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
```
33Orange, line 20:
```
```

Let's add some error messages.

src/controllers/expenseController.js
```diff
@@ -0,0 +1,82 @@
...
+  res.send(foundExpense);
+};
+
+const createExpenseController = (req, res) => {
+  const { title, userId } = req.body;
+  const user = userService.getOne(userId);
+  const isMandatoryData = title && userId;
+
+  if (!isMandatoryData || !user) {
+    res.sendStatus(400);
```
33Orange, line 35: Same here and in all other cases

src/controllers/userController.js
```diff
@@ -0,0 +1,84 @@
...
+  res.statusCode = 200;
+  res.send(users);
+};
+
+const getOneUserController = (req, res) => {
+  const { id } = req.params;
+  const foundUser = userService.getOne(id);
+
+  if (!foundUser) {
+    res.sendStatus(404);
```
33Orange, line 17: Same about the error messages

src/services/usersService.js
```diff
@@ -0,0 +1,43 @@
...
+};
+
+class UserService {
+  getAll() {
+    return users;
+  };
+  getOne(userId) {
+    return users.find(({ id }) => id === +userId) || null;
+  };
+  create(name) {
```
33Orange, line 20: Let's add some spacing between methods

```suggestion
  getAll() {
    return users;
  };
  
  getOne(userId) {
    return users.find(({ id }) => id === +userId) || null;
  };
  
  create(name) {
```

src/services/expensesService.js
```diff
@@ -0,0 +1,72 @@
...
+      from,
+      to,
+    );
+
+    return filteredExpenses;
+  };
+  getOne(expenseId) {
+    return expenses.find(({ id }) => id === +expenseId) || null;
+  };
+  create({
```
33Orange, line 34: Same about spacing

33Orange (changes requested):
Good job.

A few moments with code style and some minor issues.

PR #292, opened 2023-07-26
node_accounting-app solution

panch1811:


src/users/usersController.js
```diff
@@ -0,0 +1,84 @@
...
+  const users = userServices.getAll();
+
+  res.send(users);
+};
+
+const create = (req, res) => {
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
```
33Orange, line 15:
```
```

Let's add some messages to the errors

src/users/usersController.js
```diff
@@ -0,0 +1,84 @@
...
+  res.statusCode = 201;
+  res.send(newUser);
+};
+
+const getOne = (req, res) => {
+  const { id } = req.params;
+  const foundUser = userServices.getUserById(id);
+
+  if (!foundUser) {
+    res.sendStatus(404);
```
33Orange, line 31: Same here and in all other cases

src/expenses/expensesServices.js
```diff
@@ -0,0 +1,113 @@
...
+const create = ({
+  userId,
+  spentAt,
+  title,
+  amount,
+  category,
+  note,
+}) => {
+  const maxId
+  = expenses.length > 0 ? Math.max(...expenses.map(({ id }) => id)) : 0;
```
33Orange, line 73:
```suggestion
  const maxId = expenses.length > 0 
    ? Math.max(...expenses.map(({ id }) => id)) 
    : 0;
```

33Orange (changes requested):
Almost done

image
Let's create 3 folders instead: Routers, Controllers, Services

PR #294, opened 2023-07-26
add task solution

EugeneSherepa:


naziamoff (changes requested):
not bad, but let's now create separate directories for controllers, services, and routes, and move our files accordingly.

Bad project architecture is probably the last thing you would like to see on a real project. It makes your code terribly difficult to support

PR #295, opened 2023-07-26
add users route

kirilshorin:


naziamoff (changes requested):
not bad, but let's now create separate directories for controllers, services and routes, and move our methods and files accordingly.

Bad project architecture is probably the last thing you would like to see on the real project. It makes your code terribly difficult to support

PR #297, opened 2023-07-26
Solution

RoksolanaTylipska:


naziamoff (changes requested):
test reject

RoksolanaTylipska:
> test reject

@naziamoff Hello! Everything are passing. 
image

src/routers/users.js
```diff
@@ -0,0 +1,120 @@
...
+
+  const newUser = {
+    id: getNewId(users),
+    name,
+  };
+
+  users.push(newUser);
+  res.status(201);
+  res.json(newUser);
+});
```
33Orange, line 57: Let's create 3 layouts. Routes, Controllers, Services.

For example:

**routes/users.js**

`router.post('/', userController.createUser);`

**controller/users.js**

```
const createUser = (req, res) => {
  const { name } = req.body;

  if (!name) {
    res.status(400);
    res.send('User is not added');

    return;
  }
  const newUser = userService.createUser(name);

  res.statusCode = 201;
  res.json(newUser);
}
```

And in service, you need to add the logic of interaction with the database.
Same in all other cases

33Orange (changes requested):
Almost done.
Let's create separate directories for controllers, services, and routes

PR #298, opened 2023-07-26
Solution

MaksimKad:


src/controllers/usersController.js
```diff
@@ -0,0 +1,100 @@
...
+  const users = userService.getUsers();
+
+  res.send(users);
+};
+
+const getOne = (req, res) => {
+  const userId = Number(req.params.userId);
+
+  if (isNaN(userId)) {
+    res.sendStatus(400);
```
33Orange, line 15:
```
```
Let's add some error messages.

src/controllers/usersController.js
```diff
@@ -0,0 +1,100 @@
...
+
+    return;
+  }
+
+  const foundUser = userService.getUser(userId);
+
+  if (!foundUser) {
+    res.sendStatus(404);
```
33Orange, line 23: Same here and in all other cases

src/controllers/usersController.js
```diff
@@ -0,0 +1,100 @@
...
+    res.sendStatus(400);
+
+    return;
+  }
+
+  if (!foundUser) {
+    res.sendStatus(404);
+
+    return;
+  }
```
33Orange, line 84: Let's find the user only if we have valid data.

```suggestion
  const { name } = req.body;
  const userId = Number(req.params.userId);
  const isNameValid = typeof name === 'string' && name.length > 0;

  if (!isNameValid) {
    res.sendStatus(400);

    return;
  }
 
  if (isNaN(userId)) {
    res.sendStatus(400);

    return;
  }

  const foundUser = userService.getUser(userId);
  
  if (!foundUser) {
    res.sendStatus(404);

    return;
  }
```

src/controllers/expensesController.js
```diff
@@ -0,0 +1,129 @@
...
+  const expenses = expensesService.getAll(req.query);
+
+  res.send(expenses);
+};
+
+const getOne = (req, res) => {
+  const expenseId = Number(req.params.expenseId);
+
+  if (isNaN(expenseId)) {
+    res.sendStatus(400);
```
33Orange, line 16: Same about the error messages

33Orange (changes requested):
Almost done

PR #299, opened 2023-07-26
node_accounting-app-v2

k1samehoshigaki:


src/controllers/users.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+
+const { UserService } = require('../services/users.service');
+
+const getAllUsers = (req, res) => {
+  const userService = new UserService();
```
33Orange, line 6: You need to create a `UserSerivce` once.
Use a singleton pattern if you want to create it in the class

https://www.freecodecamp.org/news/singleton-design-pattern-with-javascript/

```suggestion
```

src/controllers/users.js
```diff
@@ -0,0 +1,92 @@
...
+  const users = userService.findAll();
+
+  res.send(users);
+};
+
+const getOneUser = (req, res) => {
+  const userService = new UserService();
```
33Orange, line 13: Same here and in all other cases

src/controllers/expenses.js
```diff
@@ -0,0 +1,87 @@
+'use strict';
+
+const { ExpenseService } = require('../services/expenses.service');
+const { UserService } = require('../services/users.service');
+
+const getAllExpenses = (req, res) => {
+  const expenseService = new ExpenseService();
```
33Orange, line 7: Same with expenses

src/controllers/users.js
```diff
@@ -0,0 +1,92 @@
...
+  const user = userService.findById(userId);
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(user);
+};
```
33Orange, line 25: Did you check the swagger for responses?

You need to send status 400, when you don't have userId in this request.
You need to send status 404, when you don't have a user with provided userId.
And you need to send status 200 with the user in the payload when everything is okay.

image

```
```

Let's add some messages to the errors.

33Orange (changes requested):
Almost done.

Let's check the swagger for the responses in all our cases.
Create the `UserService` and the `ExpensesService` once as `singleton classes`.

PR #300, opened 2023-07-27
Solution

amashutikov:


src/controllers/users.js
```diff
@@ -0,0 +1,79 @@
...
+
+  res.send(users);
+};
+
+const getOne = (req, res) => {
+  const { userId } = req.params;
+  const foundUser = userService.getById(userId);
+
+  if (!foundUser) {
+    res.sendStatus(404);
```
33Orange, line 16:
```
```
Let's add some message to the error

src/controllers/users.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  res.send(foundUser);
+};
+
+const add = (req, res) => {
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
```
33Orange, line 28: Same here and in all other cases

src/controllers/users.js
```diff
@@ -0,0 +1,79 @@
...
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const { name } = req.body;
+
+  userService.update(userId, name);
+
+  res.send(foundUser);
```
33Orange, line 68: image

Send the error with status code 400, when you don't get a userId.
Send the error with status code 404, when you don't find a user.
Send the response with status code 200, when everything is fine.

src/controllers/users.js
```diff
@@ -0,0 +1,79 @@
...
+
+const getAll = (req, res) => {
+  const users = userService.getAll();
+
+  res.send(users);
+};
+
+const getOne = (req, res) => {
+  const { userId } = req.params;
+  const foundUser = userService.getById(userId);
```
33Orange, line 13: image

src/services/users.js
```diff
@@ -0,0 +1,48 @@
...
+
+const { createNewId } = require('../helpers.js');
+
+let users = [];
+
+const resetUsers = () => {
+  users = [];
+};
+const getAll = () => users;
+const getById = (id) => {
```
33Orange, line 11: Let's add some break lines between functions

```suggestion
};

const getAll = () => users;

const getById = (id) => {
```

src/services/expenses.js
```diff
@@ -0,0 +1,74 @@
...
+
+const { createNewId } = require('../helpers.js');
+
+let expenses = [];
+
+const resetExpenses = () => {
+  expenses = [];
+};
+const getAll = () => expenses;
+const getFiltered = (userId, from, to, categories) => {
```
33Orange, line 11:
```suggestion
};

const getAll = () => expenses;

const getFiltered = (userId, from, to, categories) => {
```

33Orange (changes requested):
Almost done

PR #302, opened 2023-07-27
solution

SofiaLiubchenko:


src/createServer.js
```diff
@@ -1,11 +1,23 @@
...
-  // Return the server (express app)
+  const app = express();
+
+  usersService.setInitialUsers();
+
+  app.use('/users', usersRouter);
+
+  expensesService.setInitialExpenses();
+
+  app.use('/expenses', expensesRouter);
```
33Orange, line 18:
```suggestion
  usersService.setInitialUsers();
  expensesService.setInitialExpenses();
  
  app.use('/users', usersRouter);
  app.use('/expenses', expensesRouter);
```

src/controllers/users.controller.js
```diff
@@ -0,0 +1,82 @@
...
+  res.send(users);
+};
+
+const getOne = (req, res) => {
+  const { userId } = req.params;
+
+  const foundedUser = usersService.getById(userId);
+
+  if (!foundedUser) {
+    res.sendStatus(404);
```
33Orange, line 17: Did you check the swagger for responses?

You need to send status 400, when you don't have userId in this request.
You need to send status 404, when you don't have a user with provided userId.
And you need to send status 200 with the user in the payload when everything is okay.

image

```
```

Let's add some messages to the errors.

src/controllers/users.controller.js
```diff
@@ -0,0 +1,82 @@
...
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newUser = usersService.create(name);
+
+  res.status(201);
+  res.send(newUser);
+};
```
33Orange, line 38: Same here and in all other cases

33Orange (changes requested):
Almost done

PR #306, opened 2023-08-08
Solution

MihailKostov:


src/services/expenseServices.js
```diff
@@ -0,0 +1,128 @@
...
+  ];
+
+  const requestedField = Object.keys(req);
+
+  const hasRequiredFields = requiredFields
+    .every(field => requestedField.includes(field));
+
+  const user = getUserById(req.userId);
+
+  if (!user || !hasRequiredFields || !requiredFields.length) {
```
VitaliyBodnar, line 27: The condition '!requiredFields.length' will never be true, because 'requiredFields' is a constant array with elements.

src/controllers/expensesController.js
```diff
@@ -0,0 +1,77 @@
...
+
+    return;
+  }
+
+  const newExpense = expenseServices.create(req.body);
+
+  res.status(201).send(newExpense);
+};
+
+const getOneExpense = (req, res) => {
```
VitaliyBodnar, line 25:
```suggestion
const getExpense = (req, res) => {
```

src/services/userServices.js
```diff
@@ -0,0 +1,49 @@
...
+
+const create = (name) => {
+  const maxId = users.length > 0
+    ? Math.max(...users.map(({ id }) => id))
+    : 0;
+
+  const newUser = {
+    id: maxId + 1,
+    name,
+  };
```
VitaliyBodnar, line 25: `maxId` it's a bad approach to creating a unique id, it's better to use the current Date, or use some library for making id, like 'uuid4'
check other places

VitaliyBodnar (changes requested):

maxim2310 (approved):
Great job!!!

PR #313, opened 2023-08-18
Add task solution

vladmakohon:


src/utils.js
```diff
@@ -0,0 +1,15 @@
...
+  let maxId = 0;
+
+  for (const item of items) {
+    if (item.id >= maxId) {
+      maxId = item.id;
+    }
+  }
+
+  return maxId + 1;
+};
```
anstsot, line 13: `uuid` library can be used

src/createServer.js
```diff
@@ -1,11 +1,211 @@
...
+    res.sendStatus(204);
+  });
+  // #endregion
+
+  // #region Expenses Routes
+  app.get('/expenses', (req, res) => {
+    const userId = +req.query.userId;
+    const dateFromStr = req.query.from;
+    const dateToStr = req.query.to;
+    const categories = req.query.categories;
```
anstsot, line 94:
```suggestion
    const {
      userId, from: dateFromStr, ...
    } = req.query;
```

anstsot (changes requested):

src/utils.js
vladmakohon: Sure, but I feel like getting item by id would be not the easiest thing then.

src/services/expenses.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+const { getNewId } = require('../utils');
+
+let expenses = [];
+
+function getFiltered(userIdStr, dateFromStr, dateToStr, categories) {
+  let filteredExpenses = expenses;
```
anstsot, line 8:
```suggestion
  let filteredExpenses = [...expenses];
```

anstsot (approved):

PR #315, opened 2023-08-19
solution

cebanoleksandr:


src/createServer.js
```diff
@@ -1,11 +1,229 @@
...
+    const { name } = req.body;
+
+    if (!name) {
+      res.sendStatus(400);
+
+      return;
+    }
+
+    const newUser = {
+      id: generateId(users),
```
anstsot, line 29: `uuid` library can be used

anstsot (changes requested):

And678 (approved):
Good job, approve :white_check_mark:

PR #317, opened 2023-08-20
Solution

outofs:


src/createServer.js
```diff
@@ -1,11 +1,228 @@
...
+
+    users = users.map(user => {
+      if (user.id === id) {
+        return ({
+          ...foundUser,
+        });
+      } else {
+        return user;
+      }
+    });
```
anstsot, line 94: simplofy
```suggestion
    users = users.map(user => user.id === id ? { ...foundUser } : user);
```

src/createServer.js
```diff
@@ -1,11 +1,228 @@
...
+
+      return;
+    }
+
+    res.send(foundExpense);
+  });
+
+  app.post('/expenses', (req, res) => {
+    const
+      { userId, title, spentAt, amount, category, note } = req.body;
```
anstsot, line 148:
```suggestion
    const {
      userId, title, spentAt, amount, category, note,
    } = req.body;
```

anstsot (changes requested):
Try to fix all tests and It will be better to refactor the code to look like [in this article](https://dev.to/ericchapman/nodejs-express-part-5-routes-and-controllers-55d3) and separate to files (routes and controllers)

src/createServer.js
```diff
@@ -1,11 +1,228 @@
...
+    const { name } = req.body;
+
+    if (!name) {
+      res.sendStatus(400);
+
+      return;
+    }
+
+    const newUser = {
+      id: Date.now(),
```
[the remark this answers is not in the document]
outofs: > better to use `uuid` library

Yeah, I know) But tests are failed if `id` type is not a number

src/controllers/expenses.js
```diff
@@ -0,0 +1,124 @@
...
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const foundUser = usersServises.getById(userId);
+
+  if (!foundUser) {
+    res.statusCode = 400;
+    res.send('User is not found');
```
anstsot, line 68:
```suggestion
    res.status(400).send('User is not found');
```

anstsot (approved):

PR #319, opened 2023-08-20
base version

ivanbilotserkivskiy:


src/createServer.js
```diff
@@ -1,12 +1,221 @@
...
+
+    if (!name) {
+      res.sendStatus(400);
+
+      return;
+    }
+
+    const newId = users.length > 0
+      ? Math.max(...users.map(user => user.id)) + 1
+      : 1000;
```
anstsot, line 56: you can use `uuid` library

anstsot (changes requested):
try to fix tests (add `cors` package)
It will be good to create separate files (Routes and Controllers) for `users` and `expenses`

src/services/expenses.js
```diff
@@ -0,0 +1,82 @@
...
+}
+
+function setFilteredExpenses(filteredExpenses) {
+  expenses = [...filteredExpenses];
+}
+
+function create({ userId, title, amount, category, note, spentAt }) {
+  const newId = expenses.length > 0
+    ? Math.max(...expenses.map(expense => expense.id)) + 1
+    : 0;
```
And678, line 63: Same for users, you can delete id=5, then create a new expense, and will have id=5 again, this is wrong, deleted ids should stay deleted.

src/controllers/expenses.js
```diff
@@ -0,0 +1,97 @@
...
+
+  if ((title && typeof title !== 'string')
+      || (category && typeof category !== 'string')
+      || (note && typeof note !== 'string')) {
+    res.sendStatus(400);
+  }
+
+  Object.assign(foundExpenses, {
+    ...req['body'],
+  });
```
And678, line 90: User can send arbitrary fields and they will be saved into the db.

And678 (changes requested):
Gj, but can be even better :)

anstsot (approved):

PR #321, opened 2023-08-21
add task solution

zoriana-melnychuk:


anstsot (changes requested):

src/desktop.ini
And678, line 1: Don't push random files, editor configuration, etc.

And678 (approved):
Nice, approve :white_check_mark:

PR #322, opened 2023-08-21
add task solution

irynabolbat:


src/createServer.js
```diff
@@ -1,11 +1,37 @@
...
+  getUserById(app);
+  createNewUser(app);
+  removeUser(app);
+  updateUser(app);
+
+  getAllExpenses(app);
+  getExpense(app);
+  createNewExpense(app);
+  removeExpense(app);
+  updateExpense(app);
```
anstsot, line 32: it's not the best approach to creating routes and it's hard to get used routes at a glance

src/createServer.js
```diff
@@ -1,11 +1,37 @@
...
+} = require('./expenses');
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express().use(cors());
+
+  app.users = [];
+  app.expenses = [];
```
anstsot, line 16: better to create 2 separate variables
```suggestion
  let users = [];
  let expenses = [];
```

src/generateId.js
```diff
@@ -0,0 +1,14 @@
...
+function generateId(arr) {
+  if (arr.length === 0) {
+    return 0;
+  }
+
+  const ids = arr.map(item => item.id);
+  const nextId = Math.max(...ids) + 1;
+
+  return nextId;
+}
```
anstsot, line 12: you can use `uuid` library instead of creating own function

anstsot (changes requested):

And678 (approved):
GJ, approved :white_check_mark:

PR #323, opened 2023-08-21
solution

AnatoliiLykhohodenko:


src/services/helpers.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+function getNewId(array) {
+  if (!array.length) {
+    return 0;
+  }
+
+  return Math.max(...array.map(item => item.id)) + 1;
+}
```
anstsot, line 9: `uuid` library can be used

anstsot (changes requested):

src/services/helpers.js
AnatoliiLykhohodenko: Tests require a number as an ID

anstsot (changes requested):
Let's improve a little and create `contollers` and `routes` files to get cleaner code

src/createServer.js
```diff
@@ -1,11 +1,17 @@
...
+const express = require('express');
+const expensesRoutes = require('./routes/expenses.js');
+const usersRoutes = require('./routes/users.js');
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
```
anstsot, line 9: clear `users` and `expenses` on each server creation (just reassign empty array)

src/controllers/expenses.js
```diff
@@ -0,0 +1,141 @@
...
+    const filteredExpenses = expensesServices.getFilteredExpenses(expenses, {
+      selectedUserId,
+      categories,
+      from,
+      to,
+    });
+
+    res.status(200).send(filteredExpenses);
+  };
+};
```
anstsot, line 21: can be shortened
```suggestion
const getAllExpenses = (expenses) => (req, res) => {
  const selectedUserId = +req.query.userId;
  const { categories, from, to } = req.query;

  const filteredExpenses = expensesServices.getFilteredExpenses(expenses, {
    selectedUserId,
    categories,
    from,
    to,
  });

  res.status(200).send(filteredExpenses);
};
```

anstsot (changes requested):
Looks good but the tests are failed (try to clear users and expenses)

PR #324, opened 2023-08-21
solution

Valentyna08:


src/createServer.js
```diff
@@ -1,11 +1,23 @@
 'use strict';
 
-// const express = require('express');
+const express = require('express');
+
+const { router: usersRouter } = require('./routes/users');
+const { clearUser: clearU } = require('./services/usersService');
```
anstsot, line 6: better not to minify
```suggestion
const { clearUser } = require('./services/usersService');
```

src/services/usersService.js
```diff
@@ -0,0 +1,53 @@
...
+function getUserById(userId) {
+  const foundUser = users.find(user => user.id === +userId);
+
+  return foundUser || null;
+}
+
+function addOne(name) {
+  const ids = users.map(user => user.id);
+
+  const maxId = Math.max(...ids, 0) + 1;
```
anstsot, line 18: the `uuid` library can be used

anstsot (approved):
But pay attention to my comments to avoid the same mistakes in the future

PR #327, opened 2023-08-21
feat: accounting app provided

OKhalanskyi:


anstsot (changes requested):
Great separation of the files :thumbsup:
Let's fix the tests

PR #328, opened 2023-08-21
Develop

dianadiv:


src/services/expenses.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+
+const generateId = () => {
+  if (expenses.length < 1) {
+    return 1;
+  }
+
+  const maxId = Math.max(...expenses.map(expense => expense.id));
```
And678, line 8: This will lead to issues. E.g you delete id=5, then create a new entity, and it will be assigned to an id that was deleted.

anstsot (approved):

PR #330, opened 2023-08-21
Finished task

IgorOksentyuk:


anstsot (changes requested):
It will be better to refactor the code to look like [in this article](https://dev.to/ericchapman/nodejs-express-part-5-routes-and-controllers-55d3) and separate to files (routes, controllers and services)

And678 (approved):
Gj, approved :white_check_mark:

PR #331, opened 2023-08-21
adding solution

OlhaSheliakina:


anstsot (changes requested):

And678 (approved):
Super, approved :white_check_mark:

PR #335, opened 2023-08-21
add task solution

dipinchuk:


anstsot (approved):
The tests failed because of missing `cors` dependency in `package.json`. You can add it to make the tests work here.

PR #336, opened 2023-08-21
add task solution

klmyana:


anstsot (changes requested):
It will be better to refactor the code to look like [in this article](https://dev.to/ericchapman/nodejs-express-part-5-routes-and-controllers-55d3) and separate to files (routes, controllers and services)

 The tests are failed. could you please fix them? (add `cors` dependency to `package.json`)

PR #337, opened 2023-08-21
add task solution

MarianVolytskyi:


anstsot (changes requested):
Looks great but the tests are failed. could you please fix them? (add `cors` dependency to `package.json`)

anstsot (approved):

PR #340, opened 2023-08-21
Solution

fawnique:
The first version of the code, without division into controllers and routers


anstsot (changes requested):

PR #346, opened 2023-08-22
Task done

AntonBaranskyi:


mbruhler (changes requested):
Please put the logic into separate service files and try-catch the logic.

PR #349, opened 2023-08-22
added task solution

tetianakostiuk1:
image
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,8 @@
     "supertest": "^6.3.0"
   },
   "dependencies": {
-    "express": "^4.18.2"
+    "cors": "^2.8.5",
+    "express": "^4.18.2",
+    "uuid": "^9.0.0"
   }
 }


src/routes/expensesRoutes.js
```diff
@@ -0,0 +1,16 @@
...
+const expenseController = require('../controllers/expenseController');
+
+expensesRouter.route('/')
+  .get(expenseController.getExpenses)
+  .post(expenseController.createExpense);
+
+expensesRouter.route('/:id')
+  .get(expenseController.getExpense)
+  .patch(expenseController.updateExpense)
+  .delete(expenseController.deleteExpense);
```
And678, line 14: Nice express usage

And678 (changes requested):
Very good solution.
Please fix the tests. You need to push `package.json` with `cors` package installed. This is important, especially on real projects. Currently, your `npm i` doesn't install `cors` and it does not work on my machine.

It works on yours because you have `cors` installed locally.

And678 (approved):
Nice :+1:

PR #352, opened 2023-08-23
Implement createServer function

taras-bilas:


src/createServer.js
```diff
@@ -1,11 +1,82 @@
...
+
+    const newUser = {
+      id: uuidv4,
+      name,
+    };
+
+    users.push(newUser);
+
+    res.sendStatus(201);
+    res.send(newUser);
```
ArtemTeslenko, line 32: chain like this res.status().send()

ArtemTeslenko (changes requested):

PR #353, opened 2023-08-23
Develop

AlyonaSarapina:


And678 (changes requested):
Looks good, but you need to separate your "controllers" in `controllers` and `services`. Controllers are functions that accept `req` and `res`, while service contains `users` and `expenses` arrays and functions that update them. Check Misha's videos about this.

YegorKorenkov (approved):
GJ!

PR #358, opened 2023-08-25
add task solution

NadiaKovalchuk:


src/controllers/expenses.js
```diff
@@ -0,0 +1,107 @@
...
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
+
+  if (!userId || !spentAt || !title || !amount || !category || !note) {
+    res.statusCode = 400;
+    res.send('Some fields were missed');
```
anstsot, line 38:
```suggestion
    res.status(400).send('Some fields were missed');
```

anstsot (changes requested):
Looks great but the tests are failed. could you please fix them? (add `cors` dependency to `package.json`)

PR #365, opened 2023-09-11
added solution

vladyslav10shevchenko:


src/controllers/expenses.js
```diff
@@ -0,0 +1,117 @@
...
+  const newExpense = expenseService.create(
+    user.id,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  );
+
+  res.statusCode = 201;
```
loralevitska, line 75: Use 'res.status(201).send(newExpense)' to set status code and send response in one line.

src/controllers/expenses.js
```diff
@@ -0,0 +1,117 @@
...
+
+  if (!foundedExpense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const expenseData = req.body;
+
+  expenseService.update(expenseId, expenseData);
```
loralevitska, line 105: The updated expense is not being sent back to the client. Consider sending the updated expense after it has been saved in the database.

src/controllers/expenses.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  res.send(foundedExpense);
+};
+
+const add = (req, res) => {
+  const { spentAt, title, amount, category, note, userId } = req.body;
+
+  if (!spentAt || !title || !amount || !category || !note || !userId) {
+    res.sendStatus(400);
```
loralevitska, line 53: The error status is sent without an explanation. Consider sending a message along with the status to make debugging easier.

src/controllers/users.js
```diff
@@ -0,0 +1,82 @@
...
+
+  const newUser = userService.create(name);
+
+  res.statusCode = 201;
+  res.send(newUser);
+};
+
+const remove = (req, res) => {
+  const { userId } = req.params;
+  const foundedUser = userService.getById(userId);
```
loralevitska, line 41: The userService.getById() function is called without await, which can lead to an unhandled promise rejection if the promise is rejected. You should use async/await or promise chaining to handle the promise properly.

src/controllers/users.js
```diff
@@ -0,0 +1,82 @@
...
+    res.sendStatus(400);
+
+    return;
+  }
+
+  userService.update({
+    userId,
+    name,
+  });
+  res.send(foundedUser);
```
loralevitska, line 75: After updating the user information, you should send the updated user data back in the response. Currently, the original user data before updating is being sent.

src/controllers/users.js
```diff
@@ -0,0 +1,82 @@
...
+
+const getAll = (req, res) => {
+  const users = userService.getAll();
+
+  res.send(users);
+};
+
+const getOne = (req, res) => {
+  const { userId } = req.params;
+  const foundedUser = userService.getById(userId);
```
loralevitska, line 13: The userService.getById() function is called without await, which can lead to an unhandled promise rejection if the promise is rejected. You should use async/await or promise chaining to handle the promise properly.

src/routes/expenses.js
```diff
@@ -0,0 +1,14 @@
...
+const express = require('express');
+const expenseController = require('../controllers/expenses');
+
+const expensesRouter = express.Router();
+
+expensesRouter.get('/', expenseController.getAll);
+expensesRouter.get('/:expenseId', expenseController.getOne);
+expensesRouter.post('/', expenseController.add);
+expensesRouter.delete('/:expenseId', expenseController.remove);
+expensesRouter.patch('/:expenseId', expenseController.update);
```
loralevitska, line 12: Don't forget about destructuring. Fix all cases

loralevitska (approved):
Good job!

PR #367, opened 2023-09-26
Solution

artyomwhite:


lerastarynets (changes requested):
pls separate your code into routes, controllers and services placed in different files and folders, it's bad approach to have everything in one place
also keep in mind that controller-is just a middleware between route and service, so it has to have as little logic as possible, all logic should be placed in service files

PR #373, opened 2023-10-14
Develop

Koliras:


src/routes/expenses.router.js
```diff
@@ -0,0 +1,18 @@
...
+
+expensesRouter.get('/', express.json(), controller.getAllExpenses);
+
+expensesRouter.get('/:id', express.json(), controller.getExpense);
+
+expensesRouter.post('/', express.json(), controller.post);
+
+expensesRouter.delete('/:id', express.json(), controller.removeExpense);
+
+expensesRouter.patch('/:id', express.json(), controller.updateExpense);
```
davinnchii, line 16: you can just write `express.json()` once in createServer.js, where you use this module

davinnchii (approved):
Gj

src/routes/expenses.router.js
Moroz-Dmytro: Fix this

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,137 @@
...
+  } = req.query;
+
+  const normalizedUrl = new URL('http://localhost:3000/' + req.url);
+
+  const categories = normalizedUrl.searchParams.getAll('categories');
+
+  let expenses = service.getAll();
+
+  if (userId) {
+    expenses = expenses.filter(expense => expense.userId === +userId);
```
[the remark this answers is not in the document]
Moroz-Dmytro: The same in all other places

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,137 @@
...
+
+  if (from) {
+    expenses = expenses.filter(expense => {
+      return new Date(expense.spentAt).valueOf() >= new Date(from).valueOf();
+    });
+  }
+
+  if (to) {
+    expenses = expenses.filter(expense => {
+      return new Date(expense.spentAt).valueOf() <= new Date(to).valueOf();
```
Moroz-Dmytro, line 31: The same avoid `return`

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,137 @@
...
+    );
+  }
+
+  res.statusCode = 200;
+  res.send(expenses);
+};
+
+const getExpense = (req, res) => {
+  const id = +req.params.id;
+  const searcedExpense = service.getById(id);
```
Moroz-Dmytro, line 47:
```suggestion
  const searchedExpense = service.getById(id);
```

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,137 @@
...
+
+  if (!id) {
+    res.sendStatus(400);
+
+    return;
+  }
```
Moroz-Dmytro, line 53: Better to do this before searching

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,137 @@
...
+
+  if (!searcedExpense) {
+    res.sendStatus(404);
```
Moroz-Dmytro, line 56: Move all statuses to constants in separate file

src/controllers/users.controller.js
```diff
@@ -0,0 +1,89 @@
...
+const service = require('../services/users.services');
+
+const getAllUsers = (req, res) => {
+  res.statusCode = 200;
+  res.send(service.getAll());
+};
+
+const getUser = (req, res) => {
+  const id = +req.params.id;
+  const searcedUser = service.getById(id);
```
Moroz-Dmytro, line 12:
```suggestion
  const searchedUser = service.getById(id);
```

src/controllers/users.controller.js
```diff
@@ -0,0 +1,89 @@
...
+
+  if (!id) {
+    res.statusCode(400);
+
+    return;
+  }
```
Moroz-Dmytro, line 18: Do it before searching

Moroz-Dmytro (changes requested):

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,137 @@
...
+const getAllExpenses = (req, res) => {
+  const {
+    userId,
+    from,
+    to,
+  } = req.query;
+
+  const normalizedUrl = new URL('http://localhost:3000/' + req.url);
+
+  const categories = normalizedUrl.searchParams.getAll('categories');
```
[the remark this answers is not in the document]
Koliras: @Moroz-Dmytro if I do as you say, some tests cannot be completed correctly

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,137 @@
...
+    });
+  }
+
+  if (to) {
+    expenses = expenses.filter(expense => {
+      return new Date(expense.spentAt).valueOf() <= new Date(to).valueOf();
+    });
+  }
+
+  if (categories.length) {
```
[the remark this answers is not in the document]
Koliras: `categories` is an array, so `if (categories)` will always be true

src/controllers/expenses.controller.js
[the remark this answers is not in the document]
Moroz-Dmytro: Try to do it with combination with my comment below 🙂

src/controllers/expenses.controller.js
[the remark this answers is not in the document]
Moroz-Dmytro: With this one
You can use `console.log` to see why it will work correctly 
If we will not have categories in searchParams -> `categories===undefined`
If we have `categories===[array of categories]`

Moroz-Dmytro (changes requested):

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  if (to) {
+    expenses = expenses.filter(expense =>
+      new Date(expense.spentAt).valueOf() <= new Date(to).valueOf());
+  }
+
+  if (categories) {
+    expenses = expenses.filter(
+      expense => categories.includes(expense.category)
```
Moroz-Dmytro, line 35: if there are no `categories` in search params we will get `undefined`

image

Moroz-Dmytro (approved):
LGTM 👍 
Only you should fix comment below

PR #374, opened 2023-10-15
Develop

oleksii-bieliaiev:


src/controllers/users.controller.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  res.send(foundUser);
+};
+
+const createUser = (req, res) => {
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
```
maxim2310, line 28: would be nice to add some meaningful message for error

src/controllers/users.controller.js
```diff
@@ -0,0 +1,79 @@
...
+  res.statusCode = 201;
+  res.send(newUser);
+};
+
+const removeUser = (req, res) => {
+  const { userId } = req.params;
+  const foundUser = usersService.getUserById(userId);
+
+  if (!foundUser) {
+    res.sendStatus(404);
```
maxim2310, line 44: would be nice to add some meaningful message for error

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,115 @@
...
+
+  if ('spentAt' in updateData) {
+    updateData['spentAt'] = new Date(updateData['spentAt']);
+  }
+
+  if (Object.keys(updateData).length === 0) {
+    res.sendStatus(400);
+
+    return;
+  }
```
maxim2310, line 103: why do you need to delete keys from req.body?

maxim2310 (changes requested):
Good job! Let's improve your code a bit)

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,97 @@
...
+
+  if (!foundUser) {
+    res.status(400).send('User Not Found');
+
+    return;
+  }
+
+  const newExpense = expenseService.createExpense(req.body);
+
+  res.statusCode = 201;
```
lerastarynets, line 57:
```suggestion
  res.status = 201;
```

lerastarynets (approved):
lgtm 🚀

PR #376, opened 2023-10-16
added express logic for users and expenses

AntonHerasymov:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,94 @@
...
+
+const getAll = (req, res) => {
+  const {
+    userId,
+    from,
+    to,
+    categories,
+  } = req.query;
+
+  const expenses = expensesFilter(userId, from, to, categories);
```
rostyslav48, line 15: Is there any point of destructuring this data here?

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,94 @@
...
+
+  res.send(expenses);
+};
+
+const getById = (req, res) => {
+  const { id } = req.params;
+  const searcedExpense = expensesService.getById(+id);
```
rostyslav48, line 22: it is better to check whether id exists first, because if it is not it will be undefined and there might be some troubles with it

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,94 @@
...
+
+const post = (req, res) => {
+  const { userId, ...rest } = req.body;
+
+  if (!userService.getById(userId)) {
+    return res.sendStatus(400);
+  }
+
+  const newExpense = {
+    id: Date.now(),
```
rostyslav48, line 47: I am not sure whether it is a good idea to add date as id, there are some libs that generate id. Besides you may just write the simple id generator
```
const idGenerator = (arr) => {
  if (!arr.length) {
    return 1;
  }

  const maxId = Math.max(...arr.map(({ id }) => id));

  return maxId + 1;
};
```

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,94 @@
...
+  const { id } = req.params;
+  const expense = expensesService.getById(+id);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const updateExpense = expensesService.update(expense, req.body);
```
rostyslav48, line 67: What if I send you wrong data to update the expense, it would break I suppose. You need to check req.body before implementing it

src/controllers/users.controller.js
```diff
@@ -0,0 +1,84 @@
...
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const user = {
+    id: +new Date(),
```
rostyslav48, line 33: Again, I am not sure whether it is a good practice to add date as id. In the majority of cases it would be unique, but...

rostyslav48 (changes requested):
Great Job🤩. However, there are some places where the code is unclear. Your wo as been rejected because of `expenses update` method, you need to check which data you get before updating the expense. All other comments are on your consideration.

src/utils/expensesFilter.js
```diff
@@ -0,0 +1,48 @@
...
+  categories
+) => {
+  let expenses = expensesService.getAll();
+
+  if (userId) {
+    expenses = expenses.filter(expense => expense.userId === +userId);
+  }
+
+  if (categories) {
+    expenses = expenses.filter(({ category }) => categories.includes(category));
```
Moroz-Dmytro, line 32: if there are no `categories` in search params we will get `undefined`

image

src/utils/idGenerator.js
```diff
@@ -0,0 +1,13 @@
...
+
+const idGenerator = (arr) => {
+  if (!arr.length) {
+    return 1;
+  }
+
+  const maxId = Math.max(...arr.map(({ id }) => id));
+
+  return maxId + 1;
+};
```
Moroz-Dmytro, line 11: You can use [uuid](https://www.npmjs.com/package/uuid)

src/controllers/expenses.controller.js
Moroz-Dmytro: Good point, you don't need destructuring

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,104 @@
...
+  const expenses = expensesFilter(userId, from, to, categories);
+
+  res.send(expenses);
+};
+
+const getById = (req, res) => {
+  const { id } = req.params;
+  const searcedExpense = expensesService.getById(+id);
+
+  if (!id) {
```
Moroz-Dmytro, line 25: Better to check it before searching

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,104 @@
...
+    res.sendStatus(400);
```
Moroz-Dmytro, line 26: Move all statuses to constants to separate module

src/controllers/expenses.controller.js
Moroz-Dmytro: Good point

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,104 @@
...
+  const expense = expensesService.getById(+id);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const hasAllValues = Object.values(req.body)
+    .every(value => value !== undefined);
```
Moroz-Dmytro, line 69: It's not really good check, because if user send only `{title: 'some text', amount: 2}` you will validate it as correct body

src/controllers/users.controller.js
```diff
@@ -0,0 +1,85 @@
...
+
+  res.status(201).send(user);
+};
+
+const updateUser = (req, res) => {
+  const { name } = req.body;
+  const { id } = req.params;
+  const userToUpdate = userService.update(+id, name);
+
+  if (!id) {
```
Moroz-Dmytro, line 48: Better to check this before updating

Moroz-Dmytro (changes requested):

src/utils/idGenerator.js
AntonHerasymov: image
id must be a number

src/controllers/users.controller.js
```diff
@@ -0,0 +1,87 @@
...
+const idGenerator = require('../utils/idGenerator');
+
+const getAll = (req, res) => {
+  res.send(userService.getAll());
+};
+
+const getById = (req, res) => {
+  const { id } = req.params;
+
+  const user = userService.getById(+id);
```
Moroz-Dmytro, line 14:
```suggestion
  const user = userService.getById(Number(id));
```

Better to use Number it's easier for reading

src/utils/expensesFilter.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  if (categories) {
+    if (Array.isArray(categories)) {
+      expenses = expenses
+        .filter(({ category }) => categories.includes(category));
+    } else {
+      expenses = expenses
+        .filter(({ category }) => category === categories);
+    }
```
Moroz-Dmytro, line 39: Better not to nest `if`
Would be better to create variable and transform `categories` to array
And then just use `filter` without `if/else`

Moroz-Dmytro (approved):
LGTM 👍 
Only check comments below

PR #380, opened 2023-10-17
implement mvc functionality

voievodik:


Denys14Samoilenko (approved):
Good work! 🤖

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,97 @@
...
+
+  res.send(expenseService.getAll(query));
+};
+
+const getById = (req, res) => {
+  const { id } = req.params;
+  const searchedExpense = expenseService.getById(id);
+
+  if (!searchedExpense) {
+    res.sendStatus(404);
```
Moroz-Dmytro, line 17: Move all statuses to constants to separate file

src/services/expense.service.js
```diff
@@ -0,0 +1,75 @@
+'use strict';
+
+let expenses = [];
+
+const getAll = ({ userId, categories, from, to }) => {
+  if (categories) {
+    expenses = expenses.filter((expense) => {
+      return categories.includes(expense.category);
```
Moroz-Dmytro, line 8: You can avoid using `return`
Check other places in code for same improvement

src/services/expense.service.js
```diff
@@ -0,0 +1,75 @@
...
+    });
+  }
+
+  if (userId) {
+    expenses = expenses.filter((expense) => +expense.userId === +userId);
+  }
+
+  if (from && to) {
```
Moroz-Dmytro, line 16: Better to split it to 2 different `if`s

src/services/expense.service.js
```diff
@@ -0,0 +1,75 @@
...
+    expenses = expenses.filter((expense) => {
+      return expense.spentAt >= from && expense.spentAt <= to;
+    });
+  }
+
+  return expenses;
+};
+
+const getById = (id) => {
+  return expenses.find(({ id: expensesId }) => +expensesId === +id);
```
Moroz-Dmytro, line 26:
```suggestion
  return expenses.find(({ id: expensesId }) => Number(expensesId) === Number(id));
```

Moroz-Dmytro: The same in other places

Moroz-Dmytro (changes requested):
Good work 👍 
Small comments below

src/services/expense.service.js
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable max-len */
+'use strict';
+
+let expenses = [];
+
+const getAll = ({ userId, categories, from, to }) => {
+  if (categories) {
+    expenses = expenses.filter(expense => categories.includes(expense.category));
```
Moroz-Dmytro, line 8: if there are no `categories` in search params we will get `undefined`

image

src/services/expense.service.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  if (userId) {
+    expenses = expenses.filter((expense) => +expense.userId === +userId);
```
Moroz-Dmytro, line 12:
```suggestion
    expenses = expenses.filter((expense) => Number(expense.userId) === Number(userId));
```

Moroz-Dmytro (approved):
LGTM 👍 
Only check comment below, it should be fixed

PR #381, opened 2023-10-17
expense tracking app implementation

polinavafik:


polinavafik:
дуже дивно 😃
image

polinavafik:
🤨
image

kaavka (approved):
GJ!

src/createServer.js
```diff
@@ -1,11 +1,21 @@
 'use strict';
 
-// const express = require('express');
+const express = require('express');
+const { usersRouter } = require('./routes/users.routes');
+const { expensesRouter } = require('./routes/expenses.routes');
```
Moroz-Dmytro, line 5: You folder is named `Routes` not `routes` which is why your tests can be failed
Better to name it in lowercase

src/Controllers/expenses.controller.js
```diff
@@ -0,0 +1,135 @@
...
+
+const getExpenses = (req, res) => {
+  const {
+    userId,
+    from,
+    to,
+  } = req.query;
+
+  const normalizedUrl = new URL('http://localhost:3000/' + req.url);
+  const categories = normalizedUrl.searchParams.getAll('categories');
```
Moroz-Dmytro, line 14:
```suggestion
    to,
    categories,
  } = req.query;
```

src/Controllers/expenses.controller.js
```diff
@@ -0,0 +1,135 @@
...
+
+  let expenses = service.getAll();
+
+  if (userId) {
+    expenses = expenses.filter(
+      expense => expense.userId === +userId
+    );
+  }
+
+  if (categories.length) {
```
Moroz-Dmytro, line 24:
```suggestion
  if (categories) {
```

src/Controllers/expenses.controller.js
```diff
@@ -0,0 +1,135 @@
...
+  } = req.query;
+
+  const normalizedUrl = new URL('http://localhost:3000/' + req.url);
+  const categories = normalizedUrl.searchParams.getAll('categories');
+
+  let expenses = service.getAll();
+
+  if (userId) {
+    expenses = expenses.filter(
+      expense => expense.userId === +userId
```
Moroz-Dmytro, line 20:
```suggestion
      expense => expense.userId === Number(userId)
```

Moroz-Dmytro: The same in all other places, it is easier for reading 🙂

src/Controllers/expenses.controller.js
```diff
@@ -0,0 +1,135 @@
...
+
+const getOneExpense = (req, res) => {
+  const id = +req.params.id;
+  const expense = service.getById(id);
+
+  if (!id) {
+    res.sendStatus(400);
+
+    return;
+  }
```
Moroz-Dmytro, line 85: You can do it before `const expense = service.getById(id);` to avoid useless calculations

src/Controllers/expenses.controller.js
```diff
@@ -0,0 +1,135 @@
...
+
+  if (!expense) {
+    res.sendStatus(404);
```
Moroz-Dmytro, line 88: Move all constants(status codes) to separate file

Moroz-Dmytro (changes requested):
If your tests have failed -> you can open "Details" here

image

Here you can find what was wrong

image

src/Controllers/expenses.controller.js
polinavafik: when i did it this way, 4 tests were failing, but I see now
it was because of  ` if (categories.length) `
😀😀😀😀 (😭💀)

src/Controllers/expenses.controller.js
polinavafik: i had it this way before, but switched to + to make is more simple)))
i guess i should bring everything back))

Moroz-Dmytro (approved):
👍 
But check comment below, it should be fixed

PR #384, opened 2023-10-17
create server

kaavka:


src/controllers/expensesController.js
```diff
@@ -0,0 +1,138 @@
...
+    {
+      ...req.query,
+      userId,
+    },
+  );
+
+  res.send(expenses);
+}
+
+function getOne(req, res) {
```
Moroz-Dmytro, line 40:
```suggestion
function getById(req, res) {
```

src/controllers/expensesController.js
```diff
@@ -0,0 +1,138 @@
...
+  const dataToPost = {
+    ...req.body,
+  };
+
+  if (
+    !getUserById(dataToPost.userId)
+    || !dataToPost.spentAt
+    || !dataToPost.title
+    || !dataToPost.amount
+    || !dataToPost.category
```
Moroz-Dmytro, line 70: Create function for this check

src/controllers/usersController.js
```diff
@@ -0,0 +1,117 @@
...
+const { validateId } = require('../utils/validateId');
+const { prepareIdFromReq } = require('../utils/prepareIdFromReq');
+
+function get(req, res) {
+  const users = getAllUsers();
+
+  res.send(users);
+}
+
+function getOne(req, res) {
```
Moroz-Dmytro, line 25:
```suggestion
function getById(req, res) {
```

src/services/expensesService.js
```diff
@@ -0,0 +1,76 @@
...
+}) {
+  let filteredExpenses = expenses;
+
+  if (queryUserId || queryCategories.length || queryFrom || queryTo) {
+    filteredExpenses = expenses.filter(({ userId, category, spentAt }) => {
+      if (queryUserId && queryUserId !== userId) {
+        return false;
+      }
+
+      if (queryCategories.length && !queryCategories.includes(category)) {
```
Moroz-Dmytro, line 26: if there are no `categories` in search params we will get `undefined`

image

Moroz-Dmytro (changes requested):
Check in `Details` what is wrong with tests action
Need to fix this

Moroz-Dmytro (approved):

PR #386, opened 2023-10-17
Develop

cty2802:


src/expenses/services/expense.service.js
```diff
@@ -0,0 +1,109 @@
...
+
+  expenses.push(expense);
+
+  return expense;
+};
+
+const update = (id, userId, spentAt, title, amount, category, note) => {
+  const expense = getById(+id);
+
+  if (userId !== undefined) {
```
Yevheniia-Sid, line 71: I think you don't need compare with undefined you can write (!userId)

Yevheniia-Sid (approved):

src/expenses/services/expense.service.js
cty2802: If you write `if (userId){} ` it would give the same result for undefined, null, and 0, but it should return true for undefined only. If userId is 0 or null, then this property should be updated. If userId is undefined - it shouldn't.

Moroz-Dmytro (changes requested):

src/expenses/services/expense.service.js
```diff
@@ -0,0 +1,85 @@
...
+
+const create = (body) => {
+  const expense = {
+    id: helper.getId(expenses),
+    userId: Number(body.userId),
+    spentAt: body.spentAt ? new Date(body.spentAt) : undefined,
+    title: body.title,
+    amount: Number(body.amount),
+    category: body.category,
+    note: body.note,
```
Moroz-Dmytro, line 58: Maybe better to check if types are correct and return error if not, than to transform them

src/users/services/user.service.js
```diff
@@ -0,0 +1,49 @@
...
+  return user;
+};
+
+const remove = (id) => {
+  users = users.filter(user => user.id !== id);
+};
+
+module.exports = {
+  getAll,
+  getById: getById,
```
Moroz-Dmytro, line 44:
```suggestion
  getById,
```

tests/expense.test.js
```diff
@@ -10,6 +12,8 @@ describe('Expense', () => {
   beforeEach(() => {
     server = createServer();
     api = supertest(server);
+    userService.clearState();
+    expenseService.clearState();
```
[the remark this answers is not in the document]
Moroz-Dmytro: So delete these changes now

tests/user.test.js
```diff
@@ -10,6 +11,7 @@ describe('User', () => {
   beforeEach(() => {
     server = createServer();
     api = supertest(server);
+    userService.clearState();
```
Moroz-Dmytro, line 14: The same here, delete these changes

Moroz-Dmytro (changes requested):
Good work 🔥 
But some comments below

src/expenses/controllers/expense.controller.js
```diff
@@ -0,0 +1,101 @@
...
+
+    return;
+  }
+
+  if (helper.isExpense(
+    userId,
+    spentAt,
+    title,
+    amount,
+    category)) {
```
Moroz-Dmytro, line 67: Better to pass object inside instead of all these values as a separate arguments

Moroz-Dmytro (approved):

PR #389, opened 2023-10-17
implemented users and expenses collections with 5 endpoints

Yevheniia-Sid:


voronine (approved):

src/createServer.js
```diff
@@ -1,11 +1,30 @@
...
+  app.get('/users/:id', UsersController.getById);
+  app.post('/users', express.json(), UsersController.create);
+  app.delete('/users/:id', UsersController.remove);
+  app.patch('/users/:id', express.json(), UsersController.update);
+
+  app.get('/expenses', ExpensesController.getAll);
+  app.get('/expenses/:id', ExpensesController.getById);
+  app.post('/expenses', express.json(), ExpensesController.create);
+  app.delete('/expenses/:id', ExpensesController.remove);
+  app.patch('/expenses/:id', express.json(), ExpensesController.update);
```
Moroz-Dmytro, line 25: Create routers

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,146 @@
...
+  } = req.query;
+
+  if (userId) {
+    expenses = expenses
+      .filter(expense => expense.userId === +userId);
+  }
+
+  if (categories) {
+    expenses = expenses
+      .filter(expense => categories.includes(expense.category));
```
Moroz-Dmytro, line 22: if there are no `categories` in search params we will get `undefined`

image

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,146 @@
...
+  const {
+    userId,
+    categories,
+    from,
+    to,
+  } = req.query;
+
+  if (userId) {
+    expenses = expenses
+      .filter(expense => expense.userId === +userId);
```
Moroz-Dmytro, line 17:
```suggestion
      .filter(expense => expense.userId === Number(userId));
```

The same in all other places

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,146 @@
...
+  const { id } = req.params;
+  const { userId, spentAt, title, amount, category, note } = req.body;
+
+  if (!ExpenseService.getById(+id)) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  if (!userId && !spentAt && !title && !amount && !category && !note) {
```
Moroz-Dmytro, line 117: But user can update only title for example and not send all other fields(because it's `patch`)
But in your case all fields are mandatory

src/services/expenses.service.js
```diff
@@ -0,0 +1,87 @@
...
+
+  if (!expense) {
+    return;
+  }
+
+  for (const key in fields) {
+    if (fields[key]) {
+      expense[key] = fields[key];
+    }
+  }
```
Moroz-Dmytro, line 71: Just use `Object.assign`

Moroz-Dmytro (changes requested):

PR #391, opened 2023-10-17
Add requests manager

neovendano:


Orpheus29 (approved):
Good job!

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,133 @@
+'use strict';
+
+const service = require('../services/expenses.services');
+const userService = require('../services/users.services');
+
+const getExpense = (req, res) => {
+  const id = Number(req.params.id);
+  const expense = service.getById(id);
+
+  if (!id) {
```
Moroz-Dmytro, line 10: Better to do this check before searching expense

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,133 @@
...
+
+  if (userId) {
+    expenses = expenses.filter(
+      expense => expense.userId === Number(userId)
+    );
+  }
+
+  if (categories) {
+    expenses = expenses.filter(
+      expense => categories.includes(expense.category)
```
Moroz-Dmytro, line 44: if there are no `categories` in search params we will get `undefined`

image

Moroz-Dmytro (changes requested):

PR #395, opened 2023-10-18
feat: add task solution

FrontKid:


OlhaMomot (approved):
Nice! 🔥

src/services/userService.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  getAll() {
+    return this.users;
+  }
+
+  getById(id) {
+    const foundUser = this.users.find(user => user.id === +id);
+
+    return foundUser || null;
```
Moroz-Dmytro, line 19: The same comment as above

src/services/userService.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  createUser(name) {
+    const newId = this.users.length
+      ? Math.max(...this.users.map(user => user.id))
+      : 0;
```
Moroz-Dmytro, line 25: The same as above
DRY

Moroz-Dmytro (changes requested):

PR #397, opened 2023-10-18
add task solution

dpidlutska:


KatOlista (approved):
good job!

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,125 @@
+'use strict';
+
+const expensesService = require('../services/expenses.service');
+const usersSevice = require('../services/users.service');
+
+const get = (req, res) => {
```
Moroz-Dmytro, line 6:
```suggestion
const getAll = (req, res) => {
```

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,125 @@
...
+    userId,
+    categories,
+    from,
+    to,
+  } = req.query;
+
+  let expenses = expensesService.getAllExpenses();
+
+  if (userId) {
+    expenses = expenses.filter(expense => expense.userId === +(userId));
```
Moroz-Dmytro, line 17:
```suggestion
    expenses = expenses.filter(expense => expense.userId === Number(userId));
```

Moroz-Dmytro: The same in all other places

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  if (categories) {
+    expenses = expenses
+      .filter(expense => categories.includes(expense.category));
```
Moroz-Dmytro, line 22: if there are no `categories` in search params we will get `undefined`

image

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,125 @@
...
+    const date = new Date(to);
+
+    expenses = expenses
+      .filter(expense => new Date(expense.spentAt) < date);
+  }
+
+  res.send(expenses);
+};
+
+const getOne = (req, res) => {
```
Moroz-Dmytro, line 42:
```suggestion
const getById = (req, res) => {
```

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,125 @@
...
+  const { id } = req.params;
+
+  const expense = expensesService.getExpenseById(id);
+
+  if (!expense) {
+    res.sendStatus(404);
```
Moroz-Dmytro, line 48: Move all statuses to constants to separate module

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,125 @@
...
+
+const post = (req, res) => {
+  const {
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
```
Moroz-Dmytro, line 64: Looks like destructuring is useless here

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,125 @@
...
+    return;
+  }
+
+  const newExpense = expensesService.addExpense(
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note
```
Moroz-Dmytro, line 78: You can pass just all object as a prop

src/controllers/users.controller.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+const usersSevice = require('../services/users.service');
+
+const get = (req, res) => {
```
Moroz-Dmytro, line 5:
```suggestion
const getAll = (req, res) => {
```

src/controllers/users.controller.js
```diff
@@ -0,0 +1,79 @@
...
+  res.send(usersSevice.getAllUsers());
+};
+
+const getOne = (req, res) => {
```
Moroz-Dmytro, line 9:
```suggestion
const getById = (req, res) => {
```

Moroz-Dmytro (changes requested):

PR #398, opened 2023-10-18
add task solution

ydratsevska:


src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,120 @@
...
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  };
+
+  expensesService.addExpense(newExpense);
+
+  res.statusCode = 201;
```
vserhiichuk, line 57: You can create a new variable for all statusCode and use it, i think

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,120 @@
...
+
+  const expense = expensesService.getExpenseById(id);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  Object.assign(expense, expenseToUpdate);
```
vserhiichuk, line 109: I think this part of code refers to expensesService

vserhiichuk (approved):
It looks good

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,120 @@
...
+const expensesService = require('../services/expensesService');
+const usersService = require('../services/usersService');
+
+const getAllExpenses = (req, res) => {
+  const { userId, from, to, categories } = req.query;
+
+  let expenses = expensesService.getAllExpenses();
+
+  if (userId) {
+    expenses = expenses.filter((e) => e.userId === +userId);
```
tiserett, line 12: call it 'el' at least, or 'exp', not just 'e' 

**Fix it everywhere**

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,120 @@
...
+    expenses = expenses.filter((e) => e.spentAt <= to);
+  }
+
+  res.send(expenses);
+};
+
+const createExpense = (req, res) => {
+  const { userId, spentAt, title, amount, category, note } = req.body;
+
+  const idForUser = +userId;
```
tiserett, line 35: 'formattedId'

use Number instead of +, for better readability

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,120 @@
...
+
+  const userById = usersService.getUserById(idForUser);
```
tiserett, line 37: just call it user

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,120 @@
...
+  };
+
+  expensesService.addExpense(newExpense);
+
+  res.statusCode = 201;
+  res.send(newExpense);
+};
+
+const getExpenseById = (req, res) => {
+  const id = +req.params.id;
```
tiserett, line 62: desctructure id from params

src/controllers/expensesControllers.js
tiserett: agreed, better create constant for statuses

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,120 @@
...
+  expensesService.removeExpenseById(id);
+
+  res.sendStatus(204);
+};
+
+const updateExpense = (req, res) => {
+  const id = +req.params.id;
+  const expenseToUpdate = req.body;
+
+  if (!req.body) {
```
tiserett, line 95: destructure body from req and use it

src/controllers/expensesControllers.js
tiserett: agreed, move to service

src/controllers/usersControllers.js
```diff
@@ -0,0 +1,92 @@
...
+
+  const userById = usersService.getUserById(userId);
+
+  if (!userById) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  Object.assign(userById, { name });
```
tiserett, line 66: move to service

src/controllers/usersControllers.js
```diff
@@ -0,0 +1,92 @@
...
+  const newUser = {
+    id: +new Date(),
+    name,
+  };
+
+  const user = usersService.addUser(newUser);
+
+  res.statusCode = 201;
+
+  res.send(user);
```
tiserett, line 43: remove gap here

src/services/expensesService.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+let expenses = [];
+
+const clearExpenses = () => (expenses = []);
```
tiserett, line 5: add curly braces here, in case you ever want to do more actions here

src/services/expensesService.js
```diff
@@ -0,0 +1,31 @@
...
+};
+
+const getExpenseById = (id) => {
+  return expenses.find(expense => expense.id === id);
+};
+
+const addExpense = (expense) => {
+  expenses.push(expense);
+
+  return expense;
```
tiserett, line 18: you send it back here, but you don't use it on the controller, why?

src/services/usersService.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+let users = [];
+
+const clear = () => (users = []);
+const getAllUsers = () => {
```
tiserett, line 6: add gap between functions

tiserett (changes requested):
Whenever you name function inside of controller/service - think twice of it's name

Why call function ```getExpenseById ``` inside of ```expensesService```? 
We call it by using ```expensesService.getExpenseById```, it's weird, way better approach is to name it as ```getById```, when we gonna call it, it will be like ```expensesService.getById``` - since we call it from ```expensesService ```- everyone understands that we are getting ```expense```, same with ```clear```, ```remove```, ```create```, etc.

PR #400, opened 2023-10-18
created expenses + users routers, controllers, services

edgergard:


src/controllers/expensesController.js
```diff
@@ -0,0 +1,121 @@
...
+    userId,
+    from,
+    to,
+    categories,
+  } = req.query;
+
+  let expenses = expensesService.getExpenses();
+
+  if (userId) {
+    expenses = expenses.filter(expense => expense.userId === +userId);
```
Moroz-Dmytro, line 21:
```suggestion
    expenses = expenses.filter(expense => expense.userId === Number(userId));
```

The same in all other places, it's easier for understanding

src/controllers/expensesController.js
```diff
@@ -0,0 +1,121 @@
...
+
+  if (to) {
+    expenses = expenses.filter(
+      expense => new Date(expense.spentAt) < new Date(to)
+    );
+  }
+
+  if (categories) {
+    expenses = expenses.filter(
+      expense => categories.includes(expense.category)
```
Moroz-Dmytro, line 38: if there are no `categories` in search params we will get `undefined`

image

src/controllers/expensesController.js
```diff
@@ -0,0 +1,121 @@
...
+
+  res.send(expenses);
+};
+
+const getExpense = (req, res) => {
+  const { id } = req.params;
+  const expense = expensesService.getExpense(+id);
+
+  if (!expense) {
+    res.sendStatus(404);
```
Moroz-Dmytro, line 50: Move all statuses to constants to separate module

src/controllers/expensesController.js
```diff
@@ -0,0 +1,121 @@
...
+
+const createExpense = (req, res) => {
+  const {
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
```
Moroz-Dmytro, line 66: Useless destructuring

src/controllers/usersController.js
```diff
@@ -0,0 +1,74 @@
...
+    return;
+  }
+
+  res.status(201).send(usersService.createUser(name));
+};
+
+const deleteUser = (req, res) => {
+  const { id } = req.params;
+
+  if (!usersService.getUser(+id)) {
```
Moroz-Dmytro, line 41: Better to create constants for such logic `!usersService.getUser(+id)` and then use this variable
The same in all other places in code

src/controllers/expensesController.js
```diff
@@ -0,0 +1,121 @@
...
+    return;
+  }
+
+  const expense = expensesService.addExpense(
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
```
Moroz-Dmytro, line 80: Better to pass all object, not as separate arguments

Moroz-Dmytro (changes requested):

PR #401, opened 2023-10-18
Implemented an expense tracking app.

AndriiKuz:


src/utils/prepareIdFromReq.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+function prepareIdFromReq(req) {
+  return Number(req.url.slice(1));
```
ivannaromanovych, line 4: Here you can use just `req.params.id` instead of slicing original url

src/utils/constants.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const PORT = 3000;
```
ivannaromanovych, line 3: You don't use this variable

ivannaromanovych (approved):
LGTM 🔥

src/services/expensesService.js
```diff
@@ -0,0 +1,76 @@
...
+}) {
+  let filteredExpenses = expenses;
+
+  if (queryUserId || queryCategories.length || queryFrom || queryTo) {
+    filteredExpenses = expenses.filter(({ userId, category, spentAt }) => {
+      if (queryUserId && queryUserId !== userId) {
+        return false;
+      }
+
+      if (queryCategories.length && !queryCategories.includes(category)) {
```
Moroz-Dmytro, line 26: if there are no `categories` in search params we will get `undefined`

image

src/utils/deleteItemById.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+function deleteItemById(itemsArray, itemId) {
+  const itemIndex = itemsArray.findIndex(({ id }) => id === itemId);
+
+  itemsArray.splice(itemIndex, 1);
```
Moroz-Dmytro, line 6: Better to use filter

Moroz-Dmytro (changes requested):

PR #402, opened 2023-10-18
initial commit

maksym-mishchanchuk:


annak1104 (approved):
Good job!
But you should create separate folders: controllers (users.controllers, expenses.controllers), routes (users.routes, expenses.routes), and services (similarly) and organize the logic and navigation in these files.

Moroz-Dmytro (changes requested):
You need to fix tests, also split everything to routers, controllers and services

PR #404, opened 2023-10-18
one test wont pass idk why

szymonpachucki:


src/controllers/expenses.js
```diff
@@ -0,0 +1,154 @@
...
+    } = req.query;
+
+    const id = +userId;
+
+    const foundUser = getById(users, id);
+
+    if (foundUser) {
+      const expensesFilteredByUser = getExpenseByUser(expenses, id);
+
+      if (category) {
```
pawel-peksa, line 30: The test is failing because we are not fulfilling the requirements regarding sending the response with the specified category.
This is the pseudo code of the failing test:
1. Create user 
2. Add an expense for that user with a "Electronics" category
3. Add an expense for that user with a "Food" category
4. Get expenses from "Electronics" category for that user.

This request should return only the expense from "Electronics" category, instead we get both expenses from the server (with "Food" and "Electronics" category)

pawel-peksa (changes requested):

choeqq (changes requested):
Waiting for any updates @szymonpachucki 👀

choeqq (changes requested):
Waiting for any updates @szymonpachucki 👀

PR #411, opened 2023-10-19
node_accounting-app-v2

annak1104:


src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,92 @@
...
+const getAll = (req, res) => {
+  res.send(expensesService.getExpenses(req.query));
+};
+
+const getExpense = (req, res) => {
+  const { id } = req.params;
+  const expense = expensesService.getById(id);
+
+  if (!expense) {
+    res.sendStatus(404);
```
Moroz-Dmytro, line 15: Move all statuses to constants to separate module

src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,92 @@
...
+
+const addExpense = (req, res) => {
+  const {
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
```
Moroz-Dmytro, line 31: Useless destructuring

src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,92 @@
...
+
+  expensesService.add(expense);
+
+  res.statusCode = 201;
+  res.send(expense);
+};
+
+const deleteExpense = (req, res) => {
+  const { id } = req.params;
+  const isExpense = expensesService.getById(+id);
```
Moroz-Dmytro, line 57:
```suggestion
  const isExpense = expensesService.getById(Number(id));
```

The same in all other places, it's easier to understand

Moroz-Dmytro, line 57: It's not boolean value, it shouldn't start with `is`

src/controllers/users.controllers.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+const userService = require('../services/users.service');
+
+const getAll = (req, res) => {
+  res.send(userService.getAll());
+};
+
+const getUser = (req, res) => {
```
Moroz-Dmytro, line 9:
```suggestion
const getUserById = (req, res) => {
```

src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+
+const expensesService = require('../services/expenses.service');
+const usersService = require('../services/users.service');
+
+const getAll = (req, res) => {
+  res.send(expensesService.getExpenses(req.query));
+};
+
+const getExpense = (req, res) => {
```
Moroz-Dmytro, line 10:
```suggestion
const getExpenseById = (req, res) => {
```

src/services/expenses.service.js
```diff
@@ -0,0 +1,59 @@
...
+let expenses = [];
+
+const getExpenses = ({ userId, categories, from, to }) => {
+  if (userId) {
+    expenses = expenses.filter(expense => +expense.userId === +userId);
+  }
+
+  if (categories && categories.length) {
+    expenses = expenses
+      .filter(expense => categories.includes(expense.category));
```
Moroz-Dmytro, line 12: if there are no `categories` in search params we will get `undefined`

image

Moroz-Dmytro (changes requested):

PR #417, opened 2023-10-22
solution

emil-owczarek:


src/createServer.js
```diff
@@ -1,11 +1,190 @@
...
+    if (!user) {
+      return res.sendStatus(404);
+    }
+
+    if (req.body.name) {
+      user.name = req.body.name;
+    }
+
+    res.status(200).send(user);
+  });
```
choeqq, line 67: how about creating a router for those routes? then you could separatem them into 2 files and only import it in create server function 😄

src/createServer.js
```diff
@@ -1,11 +1,190 @@
...
+    if (category) {
+      expense.category = category;
+    }
+
+    if (note) {
+      expense.note = note;
+    }
+
+    res.status(200).send(expense);
+  });
```
choeqq, line 185: how about creating a router for those routes? then you could separatem them into 2 files and only import it in create server function 😄

choeqq (changes requested):
Also please consider creating services (to interact with your 'database') for each of routers 😄

choeqq (changes requested):
Waiting for any updates @emil-owczarek 👀

pawel-peksa (approved):
Good job! 💪 
No comments :)

PR #418, opened 2023-10-22
Solution

Nightycake:


src/controllers/expenses.controller.js
choeqq, line 1: please make sure to provide more descriptive names, despite the fact that you have controller name imported

src/services/user.service.js
```diff
@@ -0,0 +1,48 @@
...
+const getAllUsers = () => {
+  return users;
+};
+
+const getUserById = (id) => {
+  return users.find(user => user.id === +id) || null;
+};
+
+const createUser = (name) => {
+  const id = +(new Date());
```
choeqq, line 14: I think you could also just use `users.length` as index 😄

src/services/expenses.service.js
```diff
@@ -0,0 +1,95 @@
...
+
+const createExpense = (
+  userId,
+  spentAt,
+  title,
+  amount,
+  category,
+  note,
+) => {
+  const id = +(new Date());
```
choeqq, line 55: I think you could also just use `expenses.length` as index

choeqq (approved):

PR #422, opened 2023-10-23
Solution

Sancz0pansa:


cardinal1991 (changes requested):
`Changes should be persistent while server is working`
`You have to work inside src/createServer.js. You should create, set up and return an express app from createServer function.`
this probably means that your data should be not stored between server starts

Sancz0pansa:
> `Changes should be persistent while server is working` `You have to work inside src/createServer.js. You should create, set up and return an express app from createServer function.` this probably means that your data should be not stored between server starts

And they are not.

choeqq (changes requested):

src/services/userServices.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const users = [];
```
pawel-peksa, line 3: Tests are failing, because we are not clearinig this list.
After every start of the server in createServer try to clear both lists
`users` and `expenses`

pawel-peksa (changes requested):
Really good structure! Good job! 💪 I've added one suggestion to make all the tests passing.

pawel-peksa (changes requested):
Really good code 💪  Let's clear the script memory (users, expenses) every time we create a new server.

src/createServer.js
```diff
@@ -1,11 +1,22 @@
...
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  expenseServices.clear();
+  userServices.clear();
+
+  app.use('/', express.json(), usersRoutes);
+  app.use('/', express.json(), expensesRoutes);
```
choeqq, line 17:
```suggestion
  app.use('/users', express.json(), usersRoutes);
  app.use('/expenses', express.json(), expensesRoutes);
```

src/routes/expensesRoutes.js
```diff
@@ -0,0 +1,13 @@
...
+
+const express = require('express');
+const router = express.Router();
+const expensesController = require('../controllers/expensesController');
+
+router.get('/expenses', expensesController.getExpenses);
+router.get('/expenses/:id', expensesController.getExpenseById);
+router.post('/expenses', expensesController.createExpense);
+router.patch('/expenses/:id', expensesController.updateExpense);
+router.delete('/expenses/:id', expensesController.deleteExpense);
```
choeqq, line 11: if you'd apply my suggestion in `createServer.js` you could omit typing `/expenses` 5 times 😄

src/routes/usersRoutes.js
```diff
@@ -0,0 +1,13 @@
...
+
+const express = require('express');
+const router = express.Router();
+const usersController = require('../controllers/usersController');
+
+router.get('/users', usersController.getUsers);
+router.get('/users/:id', usersController.getUserById);
+router.post('/users', usersController.createUser);
+router.patch('/users/:id', usersController.updateUser);
+router.delete('/users/:id', usersController.deleteUser);
```
choeqq, line 11: if you'd apply my suggestion in `createServer.js` you could omit typing `/users` 5 times 😄

choeqq (changes requested):

choeqq (approved):

PR #423, opened 2023-10-23
initial commit

cardinal1991:


src/createServer.js
choeqq, line 1: Could you please divide your code into routes, services and controllers? It would be way more readable and testable 😄

choeqq (changes requested):

choeqq (approved):

PR #427, opened 2023-10-24
solution

Dolnys:


Matylda-Lewandowska (approved):
Looks good to me :)

src/endpoints/expensesControllers.js
```diff
@@ -0,0 +1,120 @@
...
+
+  if (from) {
+    expenses = expenses.filter((e) => e.spentAt >= from);
+  }
+
+  if (to) {
+    expenses = expenses.filter((e) => e.spentAt <= to);
+  }
+
+  res.send(expenses);
```
choeqq, line 29: could also send status here?

src/endpoints/expensesControllers.js
```diff
@@ -0,0 +1,120 @@
...
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  Object.assign(expense, expenseToUpdate);
+
+  res.send(expense);
```
choeqq, line 111: could also send status here?

src/endpoints/usersControllers.js
```diff
@@ -0,0 +1,92 @@
...
+
+  const user = usersService.getUserById(id);
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(user);
```
choeqq, line 22: could also send status here?

choeqq (changes requested):
I think the name endpoints should be changed to `/controllers` it confuses routes and endpoints in my opinion

Dolnys:
I changed the folder name. Do you mean it should send a status of 200? Isn't that provided by Express as the default code?

pawel-peksa (approved):
Well done 💪

PR #429, opened 2023-10-24
Add task solution

klaudia-klis:


src/controllers/users.controller.js
```diff
@@ -0,0 +1,76 @@
...
+    return;
+  }
+
+  if (!name || typeof name !== 'string') {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  res.send(userServices.updateUser(userId, name));
```
choeqq, line 67: could also send status here?

choeqq (changes requested):

pawel-peksa (approved):
Well done 🔥

PR #432, opened 2023-10-24
solution

PawelP2001:


src/controllers/users.controller.js
```diff
@@ -0,0 +1,75 @@
...
+  const { userId } = request.params;
+  const findUserById = usersServices.getUserById(userId);
+
+  if (!findUserById) {
+    response.sendStatus(404);
+
+    return;
+  }
+
+  response.send(findUserById);
```
choeqq, line 31: could also add status here?

src/controllers/users.controller.js
```diff
@@ -0,0 +1,75 @@
...
+    return;
+  }
+
+  if (!name || typeof name !== 'string') {
+    response.sendStatus(400);
+
+    return;
+  }
+
+  response.send(usersServices.updateUser(userId, name));
```
choeqq, line 66: could also add status here?

choeqq (changes requested):

choeqq (approved):
LGTM ✅

PR #434, opened 2023-10-31
new solution

yurii-krymskiy:


lerastarynets (approved):
Great job overall! Here is one tiny suggestion for the improvement
Pls move the majority of logic from controllers to services, controllers are supposed to have as least logic as possible, they r just a middleware between route and service, so pls try refactoring a little bit.

PR #435, opened 2023-11-01
An accounting app has been made!

bojkovladislav:


src/controllers/userController.js
```diff
@@ -0,0 +1,93 @@
...
+
+const getUsers = (req, res) => {
+  if (!getAllUsers()) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.status(200).send(getAllUsers());
+};
```
DarkMistyRoom, line 21: why do you need to run getAllUsers twice?

`const users = getAllUsers();`

and operate the `users` variable

src/controllers/expensesController.js
```diff
@@ -0,0 +1,112 @@
...
+const {
+  generateUniqueID,
+  checkId,
+  checkIfItemFound,
+} = require('../helpers.js');
+
+const getExpenses = (req, res) => {
+  const { userId, categories, from, to } = req.query;
+
+  if (userId || categories || (from && to)) {
```
DarkMistyRoom, line 22: looks like we need to be able to query only `from` or only `to`

DarkMistyRoom (approved):
Good job! Fix a few moments

PR #443, opened 2023-11-22
solution

OlegPopovych:
Hi

### DIFF
--- a/src/controllers/users.controller.js
+++ b/src/controllers/users.controller.js
@@ -1,0 +1,89 @@
+'use strict';
+
+const usersService = require('../services/user.service');
+
+const getAll = (req, res) => {
+  res.send(usersService.getAllUsers());
+};
+
+const getParticular = (req, res) => {
+  const { id } = req.params;
+
+  if (!id) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const user = usersService.getUserById(Number(id));
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(user);
+};
+
+const deleteParticular = (req, res) => {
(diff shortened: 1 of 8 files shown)


src/controllers/users.controller.js
```diff
@@ -0,0 +1,89 @@
+'use strict';
+
+const usersService = require('../services/user.service');
+
+const getAll = (req, res) => {
+  res.send(usersService.getAllUsers());
+};
```
StasSokolov, line 7: for now it is okay, because users is always exist, but in a real project you should handle error here, because you will take users from db etc.

src/utils/filter.expenses.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  if (categories) {
+    newExpenses = newExpenses.filter(exp => exp.category === categories);
+  }
+
+  if (from && to) {
+    newExpenses = newExpenses
+      .filter(exp => exp.spentAt > from && exp.spentAt < to);
+  }
```
StasSokolov, line 18: please use condition in one `.filter` method, because now it will invoke filter a lot of times of there is a multiple filters in query

StasSokolov (changes requested):
Almost masterpiece, let's fix only one minor comment🔥

src/controllers/users.controller.js
OlegPopovych: Yeah, I just in the middle of implementing the next app with PostgreSQL

src/utils/filter.expenses.js
OlegPopovych: Thanks, done.

StasSokolov (approved):
Looks good to me, i'm approving it🔥

PR #444, opened 2023-11-22
accounting app first commit

Sergey-Mironenko:


src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,83 @@
...
+  
+    res.send(expense);
+  };
+  
+  export const add = (req, res) => {
+    const { userId, spentAt, title, amount, category, note }  = req.body;
+  
+    if (!userId || !spentAt || !title || !amount || !category) {
+      res.statusCode(404);
+      res.end('Name is required');
```
denys-danyliuk, line 32: and if it's some other field that's not being passed?
it will be confusing for a user

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,83 @@
...
+      return;
+    }
+  
+    const newExpense = addExpense(userId, spentAt, title, amount, category, note);
+      
+    res.statusCode(201);
+  
+    res.send(newExpense);
```
denys-danyliuk, line 40: you should use try catch here, what if expense won't be created?

denys-danyliuk (changes requested):
pay attention to the linter errors, also work on your naming and overall logic
Знімок екрана 2023-11-23 о 10 57 28

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,97 @@
+const {
+    getAllExpenses,
+    getExpenseById,
+    addExpense,
+    updateExpense,
+    deleteExpense,
+  } = require('../services/expensesService');
+  
+const loadExpenses = (req, res) => {
```
SanyaBratashchuk, line 9:
```suggestion
const getExpenses = (req, res) => {

```

src/controllers/expensesControllers.js
```diff
@@ -0,0 +1,97 @@
...
+    addExpense,
+    updateExpense,
+    deleteExpense,
+  } = require('../services/expensesService');
+  
+const loadExpenses = (req, res) => {
+  res.send(getAllExpenses);
+};
+  
+const loadOneExpense = (req, res) => {
```
SanyaBratashchuk, line 13:
```suggestion
const getExpenseById = (req, res) => {

```

src/controllers/expensesControllers.js
SanyaBratashchuk: It is more common names

src/routes/expensesRoutes.js
```diff
@@ -0,0 +1,24 @@
...
+  loadExpenses,
+  loadOneExpense,
+  addOneExpense,
+  changeOneExpense,
+  removeExpense,
+} = require('../controllers/expensesControllers');
+
+const expensesRouter = express.Router();
+
+router.get('/', cors(), loadExpenses);
```
SanyaBratashchuk, line 12: where do you import cors?

SanyaBratashchuk: also router is not defined here

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #445, opened 2023-11-27
added task solution

koros-rk:
Якщо ранити тести скопом вони валяться, але при запуску тестів поодному всі тести успішно проходять
image
image

### DIFF
--- a/src/controllers/ExpenseController.js
+++ b/src/controllers/ExpenseController.js
@@ -1,0 +1,101 @@
+'use strict';
+
+const ExpenseModel = require('../models/Expense');
+const UserModel = require('../models/User');
+
+const get = (req, res) => {
+  const params = req.query;
+
+  const expenses = ExpenseModel.get(params);
+
+  res.send(expenses);
+};
+
+const getById = (req, res) => {
+  const { id } = req.params;
+
+  if (id) {
+    const expense = ExpenseModel.getById({ id: +id });
+
+    if (expense) {
+      res.send(expense);
+
+      return;
+    }
+
+    res.sendStatus(404);
+
+    return;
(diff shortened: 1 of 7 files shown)


src/controllers/ExpenseController.js
```diff
@@ -0,0 +1,101 @@
...
+
+    return;
+  }
+
+  res.sendStatus(400);
+};
+
+const remove = (req, res) => {
+  const { id } = req.params;
+  const expenseIndex = ExpenseModel.remove({ id: +id });
```
vasylmiskiv, line 63: Would be better to check if the resource exists or not. If not return the error with status code 404, otherwise delete the resource and return the status with some message (optional)

src/models/User.js
```diff
@@ -0,0 +1,57 @@
...
+
+  users.push({
+    id: nextID, name,
+  });
+
+  return getById({ id: nextID });
+};
+
+const update = ({ id, name }) => {
+  const user = getById({ id });
```
vasylmiskiv, line 28: It should check in controller in this case if the user exists or not if not return an error otherwise do update function

src/models/Expense.js
```diff
@@ -0,0 +1,80 @@
...
+
+  expenses.push({
+    id: nextID, title, amount, category, note, spentAt, userId,
+  });
+
+  return getById({ id: nextID });
+};
+
+const update = ({ id, newParams }) => {
+  const expense = getById({ id });
```
vasylmiskiv, line 49: Should check it in controller and return an error

vasylmiskiv (approved):
Overall looks fine but pay attention to the comments

PR #448, opened 2023-12-05
local tests passed

yuliabakun:


Helen-arch (approved):
Good job!

src/services/expense.service.js
```diff
@@ -0,0 +1,90 @@
...
+    const dateFrom = new Date(fromDate);
+
+    filterData = filterData.filter(ex => new Date(ex.spentAt) >= dateFrom);
+  }
+
+  if (toDate) {
+    const dateTo = new Date(toDate);
+
+    filterData = filterData.filter(ex => new Date(ex.spentAt) <= dateTo);
+  }
```
StasSokolov, line 39: please use condition in one `.filter` method, because now it will invoke filter a lot of times of there is a multiple filters in query

src/controllers/user.controller.js
```diff
@@ -0,0 +1,85 @@
+'use strict';
+
+const userService = require('../services/user.service');
+
+const get = (req, res) => {
+  res.statusCode = 200;
+  res.send(userService.getAll());
+};
```
StasSokolov, line 8: for now it is okay, because users is always exist, but in a real project you should handle error here, because you will take users from db etc.

StasSokolov (changes requested):
### Almost masterpiece, let's fix only one minor comment🔥

src/controllers/user.controller.js
yulabakun: thanks, I will handle it in the next task

StasSokolov (approved):
### Looks really good, i'm approving it🔥

PR #450, opened 2023-12-05
solution

vlevik:


src/controllers/userController.js
```diff
@@ -0,0 +1,110 @@
...
+  if (!allUsersId.includes(id)) {
+    res.statusCode = 404;
+
+    return res.send();
+  }
+
+  removeUser(id);
+
+  res.statusCode = 204;
+  res.send();
```
BetterSol, line 77: you can easily merge these two lines into one like this: `res.sendStatus(204)` (everywhere in the code)

BetterSol (approved):

src/controllers/expenseController.js
```diff
@@ -0,0 +1,147 @@
...
+  const id = +req.params.id;
+  const {
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
+
+  if (id === undefined) {
```
vasylmiskiv, line 115:
```suggestion
  if (!id) {
```

vasylmiskiv (approved):
Overlall looks fine but pay attetnion to the comments

PR #452, opened 2023-12-05
`Accounting App`

sviatikkondur:


HardCodWorker (commented):
gj!

HardCodWorker (approved):

src/services/expenses.service.js
```diff
@@ -0,0 +1,70 @@
...
+    expensesToDisplay = expensesToDisplay.filter(e => (
+      Date.parse(e.spentAt) > Date.parse(from)
+    ));
+  }
+
+  if (to) {
+    expensesToDisplay = expensesToDisplay.filter(e => (
+      Date.parse(e.spentAt) < Date.parse(to)
+    ));
+  }
```
StasSokolov, line 31: please use condition in one `.filter` method, because now it will invoke filter a lot of times of there is a multiple filters in query

StasSokolov (changes requested):
### Almost done, only one minor comment🔥

StasSokolov (approved):
### Great job, i'm approving it🔥

PR #453, opened 2023-12-05
accounting_app-v2

BetterSol:


src/createServer.js
```diff
@@ -1,11 +1,163 @@
...
+
+    if (from && to) {
+      filteredExpenses = filteredExpenses
+        .filter(e => e.spentAt > from && e.spentAt < to);
+    }
+
+    if (categories) {
+      filteredExpenses = filteredExpenses
+        .filter(e => categories.includes(e.category));
+    }
```
StasSokolov, line 97: please use condition in one `.filter` method, because now it will invoke filter a lot of times of there is a multiple filters in query

StasSokolov (changes requested):
### Almost done, let's fix only one minor comment🔥

PR #455, opened 2023-12-05
accounting app solution

o-drozzdyk:
Will create separate files for this task soon


sviatikkondur (approved):
Good job! Just move it in separate files.

PR #464, opened 2023-12-06
Solution

adioskv:


dolphinnq (approved):
Good job!

src/users/controller/user.controller.js
```diff
@@ -0,0 +1,88 @@
+'use strict';
+
+const userService = require('../services/user.service').services;
+
+function get(req, res) {
+  res.send(userService.getUsers());
+}
```
StasSokolov, line 7: for now it is okay, because users is always exist, but in a real project you should handle error here, because you will take users from db etc.

src/expenses/services/expense.service.js
```diff
@@ -0,0 +1,95 @@
...
+    filteredExpenses = filteredExpenses.filter(
+      expense => expense.spentAt > from
+    );
+  }
+
+  if (to) {
+    filteredExpenses = filteredExpenses.filter(
+      expense => expense.spentAt < to
+    );
+  }
```
StasSokolov, line 36: please use condition in one `.filter` method, because now it will invoke filter a lot of times of there is a multiple filters in query

StasSokolov (changes requested):
### Almost done, only one minor comment before approve🔥

PR #468, opened 2023-12-07
add task solution

PaninDmytro:


src/services/user.servise.js
```diff
@@ -0,0 +1,47 @@
...
+
+const getById = (id) => {
+  return users.find(user => user.id === +id) || null;
+};
+
+const create = (name) => {
+  const newId = Math.max(users.map(userId => +userId.id)) + 1;
+
+  const user = {
+    id: +newId,
```
ivaZaiets, line 17: I think operator ```+``` before ```newId``` is not necessary in the context of the user object creation, as you've already converted ```newId``` to a number when you did ```Math.max(users.map(userId => +userId.id)) + 1;```

ivaZaiets (approved):
Good job!!

PR #473, opened 2023-12-20
accounting app

dimkamg21:
accounting app

### DIFF
--- a/src/controllers/expenses.controller.js
+++ b/src/controllers/expenses.controller.js
@@ -1,0 +1,77 @@
+'use strict';
+
+const expenseService = require('../services/expenses.service');
+
+const get = (req, res) => {
+  res.send(expenseService.getExpenses());
+};
+
+const getOne = (req, res) => {
+  const { id } = req.params;
+
+  const normalizedId = parseInt(id);
+
+  const expense = expenseService.getExpenseById(normalizedId);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(expense);
+};
+
+const add = (req, res) => {
+  const expense = req.body;
+
+  if (!expense) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newExpense = expenseService.addExpense(expense);
+
+  res.send(newExpense);
+};
+
+const remove = (req, res) => {
+  const { id } = req.params;
+
+  const normalizedId = parseInt(id);
+
+  const expense = expenseService.getExpenseById(normalizedId);
+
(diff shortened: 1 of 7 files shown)


src/controllers/users.controller.js
```diff
@@ -0,0 +1,78 @@
+'use strict';
+
+const userService = require('../services/users.service');
+
+const get = (req, res) => {
+  res.send(userService.getUsers());
+};
+
+const getOne = (req, res) => {
+  const { id } = req.params;
```
DarkMistyRoom, line 10: check if id was passed

src/controllers/users.controller.js
```diff
@@ -0,0 +1,78 @@
...
+
+  const normalizedId = parseInt(id);
```
DarkMistyRoom, line 12: better do it inside the service

src/controllers/users.controller.js
```diff
@@ -0,0 +1,78 @@
...
+
+  if (!name) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newUser = userService.createUser(name);
+
+  res.send(newUser);
```
DarkMistyRoom, line 36: set appropriate status before sending

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,77 @@
...
+
+  const updatedExpense = expenseService
+    .updateExpense(normalizedId, newProperties);
+
+  res.send(updatedExpense);
+};
+
+module.exports = {
+  get, getOne, add, remove, update,
+};
```
DarkMistyRoom, line 77: the same suggestions as for `users`

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,77 @@
+'use strict';
+
+const expenseService = require('../services/expenses.service');
+
+const get = (req, res) => {
+  res.send(expenseService.getExpenses());
+};
```
DarkMistyRoom, line 7: this request can contain the parameters
image

DarkMistyRoom (changes requested):
add cors to the dependencies
image

src/controllers/expenses.controller.js
lerastarynets: not fixed

src/controllers/users.controller.js
lerastarynets: not fixed

lerastarynets (changes requested):
Almost there, pls fix all the previous comments)

PR #474, opened 2023-12-21
implement solution

kseniia-chepur:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,96 @@
...
+
+  res.statusCode = 200;
+  res.send(filteredExpenses);
+};
+
+const create = (req, res) => {
+  const { userId, spentAt, title, amount, category, note } = req.body;
+
+  if (!usersService.getById(userId) || !title) {
+    res.sendStatus(400);
```
loralevitska, line 33: Specify messages. Fix all cases 
image

loralevitska (changes requested):
Almost done) Let's improve

src/controllers/expenses.controller.js
kseniia-chepur: Done!

nazarmatsevych (approved):
Great work 🔥 🔥 🔥

PR #475, opened 2023-12-22
Solution

Viktorianeimesh:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,112 @@
...
+  };
+
+  if (userId) {
+    filteredExpenses = allExpenses.filter(
+      item => item.userId === Number(userId));
+
+    return res.send(filteredExpenses);
+  }
+
+  if (allExpenses.length === 0) {
```
loralevitska, line 24:
```suggestion
  if (!allExpenses.length) {
```

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,112 @@
...
+  return res.send(allExpenses);
+};
+
+const getOne = (req, res) => {
+  const { id } = req.params;
+
+  const item = expensesService.getExpensesById(id);
+
+  if (!item) {
+    res.sendStatus(404);
```
loralevitska, line 44: add message according to task requirement, fix all cases
image

loralevitska (changes requested):
Almost done) Let's improve

src/controllers/user.controller.js
```diff
@@ -0,0 +1,117 @@
+/* eslint-disable no-console */
+/* eslint-disable no-useless-return */
```
Andriy-Shymkiv, line 2: please do not disable this rule - `no-useless-return`
it saying that you doing return when it doesn't make any sense

src/controllers/user.controller.js
```diff
@@ -0,0 +1,117 @@
...
+
+const userService = require('../services/users.service');
+
+const get = (req, res) => {
+  const allUsers = userService.getAllUsers();
+
+  if (!allUsers) {
+    res.status(404).send('Not Found: The specified entity does not exist');
+
+    return;
```
Andriy-Shymkiv, line 13: here for example. you don't need to return. after `res.send` the function ends its call

Andriy-Shymkiv: fix it everywhere

Andriy-Shymkiv (changes requested):

maxim2310 (approved):
Good work!!!

PR #476, opened 2023-12-23
Created accounting app

Orchibald:
Test aren`t passed, but I checked through postman and everything worked


Orchibald:
> Tests aren`t passed, but I checked through postman and everything worked

src/services/expenses.service.js
```diff
@@ -0,0 +1,61 @@
...
+
+  switch (true) {
+    case userId:
+      return expenses.find(expanse => expanse.userId === userId);
+    case categories:
+      return expenses.find(expanse => expanse.categories.includes(categories));
+    case from:
+      return expenses.find(expanse => expanse.spentAt >= from);
+    case to:
+      return expenses.find(expanse => expanse.spentAt <= to);
```
Andriy-Shymkiv, line 16: you must filter the entire array of expenses and return all of them that match the criteria

src/services/expenses.service.js
```diff
@@ -0,0 +1,61 @@
...
+    default:
+      return expenses;
+  }
+};
+
+const getExpenseById = (id) => {
+  const normalizedId = parseInt(id);
+
+  return expenses.find(expense => expense.id === normalizedId) || null;
+};
```
Andriy-Shymkiv, line 26:
```suggestion
const getExpenseById = (id) => {
  return expenses.find(expense => expense.id === Number(id)) || null;
};
```

src/services/expenses.service.js
```diff
@@ -0,0 +1,61 @@
...
+  expenses.push(newExpense);
+
+  return newExpense;
+};
+
+const deleteExpense = (id) => {
+  const normalizedId = parseInt(id);
+
+  expenses = expenses.filter(expense => expense.id !== normalizedId);
+};
```
Andriy-Shymkiv, line 45: you can `return true` from this function, because now it return undefined.

src/services/user.service.js
```diff
@@ -0,0 +1,60 @@
...
+
+let users = [];
+
+const getUsers = () => users;
+
+const getUserById = (id) => {
+  const normalizedId = parseInt(id);
+
+  return users.find(user => user.id === normalizedId) || null;
+};
```
Andriy-Shymkiv, line 11: same approach as I mentioned above

Andriy-Shymkiv (changes requested):
good job! just remember that in real application with database, you should add try/catch block for all route hanlders in app

SanyaBratashchuk (approved):

PR #477, opened 2023-12-25
implemented endpoints for /users

oleksandrshtonda:


src/controllers/User.controller.js
```diff
@@ -0,0 +1,99 @@
...
+  } catch (error) {
+    console.log(error);
+    response.status(500).send('Internal Server Error');
+  }
+};
+
+const update = async(request, response) => {
+  try {
+    const { id } = request.params;
+    const updatedUser = await UserService.update(request.body.name, id);
```
l4st1m0za, line 64: Avoid using 'request.body.name' in the 'update' function. Instead, pass the whole 'request.body' as a parameter to the 'UserService.update' method. This will make the function more generic and allow you to update other properties of the user in the future without changing the controller code.

src/controllers/User.controller.js
```diff
@@ -0,0 +1,99 @@
...
+    const allUsers = await UserService.getAll();
+
+    if (!allUsers) {
+      response.status(404).send('Not Found: The specified entity does not exist');
+    }
+
+    response.statusCode = 200;
+    response.send(allUsers);
+  } catch (error) {
+    console.log(error);
```
l4st1m0za, line 18: Instead of just logging the error, provide more context or information to the client about what went wrong.

l4st1m0za (approved):
Great job! Consider comments for the future tasks!

PR #478, opened 2024-01-03
Create_accounting_app

rialleons:


src/createServer.js
```diff
@@ -1,11 +1,20 @@
...
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  app.use(cors());
+
+  app.use('/users', express.json(), userRouter);
+  app.use('/expenses', express.json(), expenseRouter);
+
+  return app;
```
DarkMistyRoom, line 17: you need to initialize empty arrays for users and expenses on each createServer running

DarkMistyRoom (changes requested):
Good job! Add initializing and try to fix the tests

src/helpers/getFilteredExpenses.js
```diff
@@ -0,0 +1,58 @@
...
+    filteredExpenses = filteredExpenses
+      .filter(expense => {
+        const expenseTime = new Date(expense.spentAt).getTime();
+
+        return expenseTime <= formattedTo;
+      });
+  }
+
+  return filteredExpenses;
+};
```
DarkMistyRoom, line 54: Sorry, I haven't mentioned this the first time. This logic looks too complicated. We don't need to rerun filtering for each parameter. We can use all conditions inside one filter method
`expenses.filter(expense => {
    if (userId && expense.userId !== +userId) {
      return false;
    }
    
   // and so on 
    `

DarkMistyRoom (changes requested):
Good job! Only one comment! I'm sure the next request will be the final

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,122 @@
...
+
+const getAll = (req, res) => {
+  const { userId, categories, from, to } = req.query;
+
+  res.send(expenseService.getAll({
+    userId,
+    categories,
+    from,
+    to,
+  }));
```
denys-danyliuk, line 14: save result of the service function to some variable and then send variable, it will increase readability of your code

src/routes/user.route.js
```diff
@@ -0,0 +1,21 @@
...
+
+userRouter.get('/', userController.getAll);
+
+userRouter.post('/', userController.create);
+
+userRouter.get('/:id', userController.getOne);
+
+userRouter.delete('/:id', userController.remove);
+
+userRouter.patch('/:id', userController.update);
```
denys-danyliuk, line 17:
```suggestion
userRouter.get('/', userController.getAll);
userRouter.post('/', userController.create);
userRouter.get('/:id', userController.getOne);
userRouter.delete('/:id', userController.remove);
userRouter.patch('/:id', userController.update);
```

denys-danyliuk (approved):

PR #479, opened 2024-01-03
solution

yevtok:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,127 @@
...
+      return item.spentAt > from && item.spentAt < to;
+    }
+  });
+
+  if (filteredExpenses.length) {
+    res.send(filteredExpenses);
+  } else {
+    res.send(allExpenses);
+  }
+};
```
roman-mirzoian, line 42:
```suggestion
const get = (req, res) => {
  const allExpenses = getAllExpenses();
  if (!allExpenses.length) {
    return res.send([]);
  }
  
  const { userId, categories, from, to } = req.query;

  const filteredExpenses = allExpenses.filter(item => {
    if (categories && userId) {
      return item.category === categories;
    }

    if (userId) {
      return item.userId === +userId;
    }

    if (from && to) {
      return item.spentAt > from && item.spentAt < to;
    }
  });
  
  return res.send(filteredExpenses.length ? filteredExpenses || allExpenses);
};
```

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,127 @@
...
+    note,
+  } = req.body;
+
+  const userToFind = getAll().find(user => user.id === +userId);
+
+  if (!title || !userToFind) {
+    res.sendStatus(400);
+
+    return;
+  }
```
roman-mirzoian, line 60:
```suggestion
  const userToFind = getAll().find(({ id }) => id === +userId);

  if (!title || !userToFind) {
    return res.status(404).send("Title or user not found.");
  }
```

roman-mirzoian (approved):

PR #480, opened 2024-01-06
result

errisi:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,156 @@
...
+  const user = getUserById(+userId);
+
+  if (!user
+    || !spentAt
+    || !title
+    || !amount
+    || !category
+    || !note
+  ) {
+    res.sendStatus(400);
```
BudnikOleksii, line 79: If you don't have some of the required fields you don't need to get a user from the database.
So it should look like: validate inputs => if ok get user => if user in the database create expense

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,156 @@
...
+  const { id } = req.params;
+  const { spentAt, title, amount, note } = req.body;
+  const expense = getExpenseById(id);
+
+  if (
+    (spentAt && typeof spentAt !== 'string')
+    || (title && typeof title !== 'string')
+    || (amount && typeof amount !== 'number')
+    || (note && typeof note !== 'string')
+  ) {
```
BudnikOleksii, line 135: The same

src/controllers/users.controller.js
```diff
@@ -0,0 +1,78 @@
...
+const add = (req, res) => {
+  const name = req.body.name;
+
+  if (!name) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newUser = addNewUser(name);
```
BudnikOleksii, line 22: Like in this controller, at first validate and only then create a new user

BudnikOleksii (changes requested):
Good work 👍 
Left a few suggestions, check them

roman-mirzoian (approved):
Well  done 👍

PR #481, opened 2024-01-07
acoounting app

Mykola-Hadupiak:
I tested everything through POSTMAN, everything works as written! and right!

tests fail due to the use of UUID (i dont wanna use prevID + 1) and the correct use of dates and other types

### DIFF
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -1,11 +1,16 @@
 'use strict';
 
-// const express = require('express');
+const express = require('express');
+const { router: userRouter } = require('./routes/user.route.js');
+const { router: expensesRouter } = require('./routes/expenses.route.js');
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  app.use('/users', express.json(), userRouter);
+  app.use('/expenses', express.json(), expensesRouter);
+
+  return app;
 }
 
 module.exports = {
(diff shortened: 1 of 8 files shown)


src/createServer.js
```diff
@@ -1,11 +1,16 @@
...
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  app.use('/users', express.json(), userRouter);
+  app.use('/expenses', express.json(), expensesRouter);
+
+  return app;
```
DarkMistyRoom, line 13: you should initialize users and expenses as empty arrays on each server running to make it work properly

src/services/expenses.service.js
```diff
@@ -0,0 +1,78 @@
...
+    const fromDate = new Date(from);
+
+    copy = copy.filter(expense => new Date(expense.spentAt) >= fromDate);
+  }
+
+  if (to) {
+    const toDate = new Date(to);
+
+    copy = copy.filter(expense => new Date(expense.spentAt) <= toDate);
+  }
```
DarkMistyRoom, line 36: so if you have all params, you will iterate 4 times?
better to do this logic due to 1 iteration only

DarkMistyRoom (changes requested):
To pass the tests using uuid, you can use a function like this:
`function generateIntId() {
  const parsedUuid = uuidParse(uuidv4());
  const buffer = Buffer.from(parsedUuid);
  const result = buffer.readUInt32BE(0);

  return result;
}`

nazarmatsevych (approved):
Awesome 🔥

PR #484, opened 2024-01-14
 solution

ArtemSydun:


src/services/userServices.js
```diff
@@ -0,0 +1,45 @@
+'use strict';
+
+let users = [];
+
+const getAllUsers = () => users;
+
+const getUserById = (id) => {
+  return users.find(user => user.id === +id) || null;
```
roman-mirzoian, line 8: image

```suggestion
  return users.find(user => user.id === +id);
```

roman-mirzoian (approved):

PR #485, opened 2024-01-16
Complete task

vlad-demchuk:
Tests failed when I ran via `npm test`, but passed if I ran each separately... I have no idea what's wrong 🙂
image
image

### DIFF
--- a/src/controllers/expense.controller.js
+++ b/src/controllers/expense.controller.js
@@ -1,0 +1,130 @@
+'use strict';
+
+const expenseService = require('../services/expense.service');
+const userService = require('../services/user.service');
+
+const get = (req, res) => {
+  const { userId, categories, from, to } = req.query;
+
+  if (userId || categories || from || to) {
+    const parsedId = parseInt(userId, 10);
+
+    if (userId && isNaN(parsedId)) {
+      res.sendStatus(400);
+
+      return;
+    }
+
+    res.send(expenseService.getByQueries({
+      userId: parsedId,
+      categories,
+      from,
+      to,
+    }));
+
+    return;
+  }
+
+  res.send(expenseService.getAll());
+};
+
+const getOne = (req, res) => {
+  const { parsedId } = req;
+
+  const expense = expenseService.getById(parsedId);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.send(expense);
+};
+
+const create = (req, res) => {
+  const { userId, spentAt, title, amount, category, note } = req.body;
+
+  const user = userService.getById(userId);
+  const isDateValid = !isNaN(Date.parse(spentAt));
+
+  if (!user
+    || !isDateValid
+    || typeof title !== 'string'
+    || typeof amount !== 'number'
+    || typeof category !== 'string'
+    || typeof note !== 'string') {
+    res.sendStatus(400);
+
+    return;
+  }
+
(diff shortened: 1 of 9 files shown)


src/services/expense.service.js
```diff
@@ -0,0 +1,104 @@
...
+    userId: userId || expense.userId,
+    spentAt: spentAt || expense.spentAt,
+    title: title || expense.title,
+    amount: amount || expense.amount,
+    category: category || expense.category,
+    note: note || expense.note,
+  });
+
+  return expense;
+};
```
SerhiiKirik, line 95: it would be easy)
```suggestion
const update(expense, req) {
  Object.assign(expense, req.body);

  return expense;
}
```

src/createServer.js
```diff
@@ -1,11 +1,17 @@
...
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  app.use('/users', express.json(), userRouter);
+
+  app.use('/expenses', express.json(), expenseRouter);
```
SerhiiKirik, line 12: you can do it like that
```suggestion

  server.use(express.json());

  app.use('/users', userRouter);
  app.use('/expenses', expenseRouter);
```

src/controllers/user.controller.js
```diff
@@ -0,0 +1,86 @@
...
+
+  if (!name || typeof name !== 'string') {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const user = userService.create(name);
+
+  res.statusCode = 201;
```
SerhiiKirik, line 34: Update in all cases
```suggestion
  res.status(201);
```

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,130 @@
...
+    parsedId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  });
+
+  res.send(updatedExpense);
+};
```
SerhiiKirik, line 122: It's extra
```suggestion
const update = (req, res) => {
  const { parsedId } = req;

  const foundExpense = expenseService.getById(parsedId);

  if (!foundExpense) {
    res.sendStatus(404);

    return;
  }
  
  const expense = { parsedId, ...req.body } 

  const updatedExpense = expenseService.update(expense);

  res.send(updatedExpense);
};
```

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,130 @@
...
+
+  const expense = expenseService.getById(parsedId);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  if ((title && typeof title !== 'string')
```
SerhiiKirik, line 103: same here

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,130 @@
...
+      userId: parsedId,
+      categories,
+      from,
+      to,
+    }));
+
+    return;
+  }
+
+  res.send(expenseService.getAll());
```
SerhiiKirik, line 28: for more readable
```suggestion
    }
    
    const expense = expenseService.getByQueries({
      userId: parsedId,
      categories,
      from,
      to,
    })

    res.send(expense);

    return;
  }
  
  const expense = expenseService.getAll();

  res.send(expense);
```

SerhiiKirik (approved):
Well done 🔥

PR #487, opened 2024-01-16
add task solution

ViktorAfk:


BoblykNazar (approved):
Good job!

src/constrolers/expenses.controler.js
```diff
@@ -0,0 +1,121 @@
...
+
+  const expense = getOneExpense(idToNum);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  Object.assign(expense, newInformation);
```
BudnikOleksii, line 110: You should update expenses in your local "database" and do it not in the controller. Controllers should only send the data to the client without manipulations with data

src/constrolers/users.controlers.js
```diff
@@ -0,0 +1,100 @@
...
+
+  const user = findUser(idToNum);
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  Object.assign(user, { name });
```
BudnikOleksii, line 67: The same

BudnikOleksii (changes requested):
Good work 👍 
Move logic for updating user/expense to the corresponding services

BudnikOleksii (approved):

PR #488, opened 2024-01-16
Develop

BoblykNazar:


mickhakk (approved):
Good work!

src/controllers/users.controller.js
```diff
@@ -0,0 +1,81 @@
...
+const updateUser = (req, res) => {
+  const userId = +req.params.id;
+  const { name } = req.body;
+  const user = userService.getById(userId);
+
+  if (!name) {
+    res.sendStatus(400);
+
+    return;
+  }
```
BudnikOleksii, line 62:
```suggestion
  if (!name) {
    res.sendStatus(400);

    return;
  }
  
  const user = userService.getById(userId);
```
If name isn't provided you don't need to query user

BudnikOleksii (approved):
Looks good 🚀

PR #495, opened 2024-01-17
Develop

VeDono:
Hello there, I have problems with tests, but I don't understand what I do wrong.
Question for mentors, can you explain reasons why many tests was failed, like:
 `updateExpense should return 404 if an expense not found` (in Insomnia I see `404 Not Foud` if `expense` does not exist), but in the test that I ran in VSCode, I always saw that and other problems (in Insomnia I caught right status codes of errors).
 
 I don't understand errors with entries in tests, when I create some entry in Insomnia - all ok, but in tests, I see errors like: 
```
- Expected  -  1
+ Received  + 42 (just many same objects of users with different id's)
```


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,170 @@
...
+const getAll = (req, res) => {
+  const {
+    userId: recivedUserId,
+    categories,
+    from,
+    to,
+  } = req.query;
+  const userId = Number(recivedUserId);
+
+  let expenses = expenseService.getAll();
```
roman-mirzoian, line 15: If `expenses.length` is 0 or less - 404 should be returned

src/controllers/user.controller.js
```diff
@@ -0,0 +1,97 @@
+'use strict';
+
+const userService = require('../services/user.service');
+
+const getAll = (_, res) => {
+  const users = userService.getAll();
```
roman-mirzoian, line 6: If `users.length` is 0 or less - 404 should be returned

roman-mirzoian (changes requested):
If you have any questions about the work, asking them in the comments to this work is the longest way to get an answer. Such questions should be asked in the QnA chat, on the QnA itself or in a personal channel.

Answering your question: currently not all possible answer options expected by the tests are covered, so it is worth reviewing the tests themselves and adding these options, as the example described the behavior for the getAll commands

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,171 @@
...
+
+  if (!expenses.length) {
+    res.send([]);
```
BudnikOleksii, line 18: If the array is empty you send manually create an empty array, that doesn't make sense

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,171 @@
...
+    expenses = expenses.filter(expense => (
+      new Date(expense.spentAt) >= fromDate
+    ));
+  }
+
+  if (to) {
+    const toDate = new Date(to);
+
+    expenses = expenses.filter(expense => new Date(expense.spentAt) <= toDate);
+  }
```
BudnikOleksii, line 47: It's a bad idea to keep logic in the controllers

src/controllers/user.controller.js
```diff
@@ -0,0 +1,103 @@
...
+
+  if (!users.length) {
+    res.send([]);
```
BudnikOleksii, line 9: The same

src/createServer.js
```diff
@@ -1,11 +1,33 @@
...
+  app.get('/users/:id', userController.getById);
+  app.post('/users', userController.create);
+  app.delete('/users/:id', userController.remove);
+  app.patch('/users/:id', userController.patch);
+
+  app.get('/expenses', expensesController.getAll);
+  app.get('/expenses/:id', expensesController.getById);
+  app.post('/expenses', expensesController.create);
+  app.delete('/expenses/:id', expensesController.remove);
+  app.patch('/expenses/:id', expensesController.patch);
```
BudnikOleksii, line 28: Don't keep all your routes in the same place, create separate router for users and expenses, and move them into separate files

src/services/expense.service.js
```diff
@@ -0,0 +1,83 @@
...
+  const expenseToUpdate = expenses.find(expense => expense.id === id);
+
+  const updates = {
+    spentAt, title, amount, category, note,
+  };
+
+  const entriesToUpdate = Object.entries(updates)
+    .filter(([_, value]) => value !== undefined);
+
+  const updatesFiltered = Object.fromEntries(entriesToUpdate);
```
BudnikOleksii, line 69:
```suggestion
  const entriesToUpdate = Object.entries(updates)
    .filter(([_, value]) => value);

  const updatesFiltered = Object.fromEntries(entriesToUpdate);
```
I think this is redundant and you don't need to compare with `undefined`

src/services/expense.service.js
```diff
@@ -0,0 +1,83 @@
...
+    amount,
+    category,
+    note,
+  } = options;
+
+  const expenseToUpdate = expenses.find(expense => expense.id === id);
+
+  const updates = {
+    spentAt, title, amount, category, note,
+  };
```
BudnikOleksii, line 64:
```suggestion
  const {
    id,
    ...updates,
  } = options;

  const expenseToUpdate = expenses.find(expense => expense.id === id);
```
You can simplify it with rest operator

BudnikOleksii (changes requested):
Good work 👍 
Lets improve your solution in accordance to comments

src/controllers/user.controller.js
VeDono: I rid that.

src/controllers/expense.controller.js
VeDono: Implement that in a separate file.

src/createServer.js
VeDono: I implement router 🤝.

src/services/expense.service.js
VeDono: Thanks for the advice.

src/services/expense.service.js
VeDono: Thanks for the amazing advice.

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,148 @@
...
+    categories,
+    from,
+    to,
+  } = req.query;
+  const userId = Number(recivedUserId);
+
+  const expenses = expenseService.getAll();
+
+  const filteredExpenses = getFilteredExpenses(
+    expenses,
```
BudnikOleksii, line 19: I meant it's a bad practice to keep any logic in the controller, it should be placed in the services. So you can filter in the `expenseService.getAll`

src/services/expense.service.js
```diff
@@ -0,0 +1,75 @@
...
+
+const { getNewId } = require('../utils/getNewId.js');
+
+let expenses = [];
+
+function resetExpenses() {
+  expenses = [];
+};
+
+const getAll = () => expenses;
```
BudnikOleksii, line 11:
```suggestion
const resetExpenses = () => {
  expenses = [];
};

const getAll = () => expenses;
```
Use the same style in your codebase

BudnikOleksii (approved):
Nice job 👍  
Approved. Just a few suggestions for the future:
- don't keep any business logic in the controllers, they should be as thin as possible, just get data and then send a response to the client.
- follow the one code style in your codebase. If you prefer arrow function, so every function should be arrow

src/services/expense.service.js
VeDono: I understud 🤝

PR #497, opened 2024-01-17
solution

emilionila:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,133 @@
...
+};
+
+const getExpenseById = (req, res) => {
+  const { expenseId } = req.params;
+  const foundExpense = expenseService.getExpenseById(expenseId);
+
+  if (!foundExpense) {
+    res.sendStatus(404);
+
+    return;
```
roman-mirzoian, line 31: In addition to setting the status, you should also add information for the user.
```suggestion
      return res.status(404).send("No expenses found.");
```

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,133 @@
...
+            && typeof filteredBody.title !== 'string')
+        || (filteredBody.hasOwnProperty('spentAt')
+            && typeof filteredBody.spentAt !== 'string')
+        || (filteredBody.hasOwnProperty('amount')
+            && typeof filteredBody.amount !== 'number')
+        || (filteredBody.hasOwnProperty('category')
+            && typeof filteredBody.category !== 'string')
+        || (filteredBody.hasOwnProperty('note')
+            && typeof filteredBody.note !== 'string')
+  ) {
```
roman-mirzoian, line 113: This looks like a very complex piece of code to understand and extend in the future.
But the same check is performed for each field, and this can be simplified by creating a helper function:
```js
function isBodyFieldInvalid(fieldName) {
  return filteredBody.hasOwnProperty(fieldName) && typeof filteredBody.title !== 'string');
}
...
if(isBodyFieldInvalid('title') || 
   isBodyFieldInvalid('spentAt') || 
   isBodyFieldInvalid('amount') || 
   isBodyFieldInvalid('category') ||
   isBodyFieldInvalid('note')) {
  ...
}
```
It's not a perfect solution, but it's much easier to understand.

src/services/expenses.service.js
```diff
@@ -0,0 +1,82 @@
...
+      .filter(expense => expense.spentAt >= from && expense.spentAt <= to);
+  }
+
+  return filteredExpenses;
+};
+
+const getExpenseById = (expenseId) => {
+  const foundExpense = expenses.find(expense => expense.id === +expenseId);
+
+  return foundExpense || null;
```
roman-mirzoian, line 30: image

```suggestion
  return expenses.find(expense => expense.id === +expenseId);
```

roman-mirzoian (approved):

PR #502, opened 2024-01-19
Solution

YuriySobakar:


src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+  });
+
+  app.delete('/users/:id', (req, res) => {
+    const { id } = req.params;
+
+    const usersLength = getAllUsers().length;
+
+    removeUser(id);
+
+    if (getAllUsers().length === usersLength) {
```
roman-mirzoian, line 78: First, you need to find out if such a user exists, and only then delete it, not the other way around

src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+  });
+
+  app.delete('/expences/:id', (req, res) => {
+    const { id } = req.params;
+
+    const expenceLength = getAllExpences().length;
+
+    removeExpence(id);
+
+    if (getAllExpences().length === expenceLength) {
```
roman-mirzoian, line 159: First, you need to find out if such an expense exists, and only then delete it, and not the other way around

src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+
+    res.sendStatus(204);
+  });
+
+  app.patch('/expences/:id', express.json(), (req, res) => {
+    const { id } = req.params;
+    const { userId, spentAt, title, amount, category, note } = req.body;
+
+    if (!getExpenceById(id)) {
+      res.sendStatus(400);
```
roman-mirzoian, line 173:
```suggestion
      res.sendStatus(404);
```

roman-mirzoian (changes requested):
Be attentive to the logic of writing the application and fix a non-working test.

roman-mirzoian (approved):
The end of studies does not mean that you can relax 😃 
Be more careful when performing tasks.

image

PR #506, opened 2024-01-28
solution

Gosyav:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,162 @@
...
+      return res
+        .send('No category')
+        .status(400);
+    }
+
+    case 'No expense to update': {
+      return res
+        .send('No expense to update')
+        .status(400);
+    }
```
DarkMistyRoom, line 60: would be better to check if all params provided in the controller and call the service method only if everything is ok

src/services/expense.service.js
```diff
@@ -0,0 +1,133 @@
...
+    return userIdMatch && categoriesMatch && dateMatch;
+  });
+
+  return filteredExpenses;
+};
+
+const post = (userId, spentAt, title, amount, category, note) => {
+  if (!userId) {
+    return 'No userId';
+  }
```
DarkMistyRoom, line 29: you should also check if the user exists in the `user` array

src/services/expense.service.js
```diff
@@ -0,0 +1,133 @@
+'use strict';
+
+let expenses = [];
+let idState = 0;
+
+const get = (userId, categories, from, to) => {
+  const filteredExpenses = expenses.filter(expense => {
+    const userIdMatch = userId.length
+      ? userId === expense.id
```
DarkMistyRoom, line 9: userId and expense.id are different things

DarkMistyRoom (changes requested):
Try to send requests to your server from postman and debug why the test are failed
image

anastasiiavorobiova (approved):
Great work!

PR #508, opened 2024-01-29
Develop

mosowiecka:


mbruhler (approved):
I approve, @Barosz30 suggestion is correct but AFAIK you weren't introduced to middlewares (services) on the lecture.

https://mate.academy/pl/learn/nodejs-pl/nodejs-express-pl?section=video&videoId=5463

There is no mention about using services/middlewares

BUT! I think this is very important and you should be introduced to that.

App works great, good job!

PR #509, opened 2024-01-29
solution

Barosz30:


aureliagorzna (approved):

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,131 @@
+'use strict';
+
+const expenseService = require('../services/expense.service');
+const { getAll } = require('../services/user.service');
+
+const get = (req, res) => {
```
Centmsn, line 6: There is a lot of bussiness logic happening in the controller itself. It should be extracted to the related service.

src/createServer.js
```diff
@@ -1,11 +1,24 @@
...
 function createServer() {
   // Use express to create a server
   // Add a routes to the server
   // Return the server (express app)
+  const app = express();
+
+  resetUsers();
+  resetExpenses();
+
+  app.use('/users', express.json(), userRouter);
```
Centmsn, line 18: `express.json()` can be used for all routes: `app.use(express.json());`. Make sure to place it before routes.
After this changed is applied, it can be removed from the specific routes.

Centmsn (changes requested):
Looks good 👍 

I left just a few comments to slightly improve the solution

mbruhler (approved):
Good job approved!

PR #510, opened 2024-01-29
instructions unclear. I had to guess what to do

aureliagorzna:


bertsu02 (approved):

main.js
```diff
@@ -1,9 +1,17 @@
...
-  });
+app.use(express.json());
+
+const usersRouter = require('./src/routers/usersRouter');
+const expensesRouter = require('./src/routers/expensesRouter');
+
+app.use('/users', usersRouter);
+app.use('/expenses', expensesRouter);
+
+app.listen(process.env.PORT || 3000, () => {
```
danon321, line 14: You duplicate this code in line 16 too so its better way to store this value in variable like const PORT = process.env.PORT || 3000

danon321 (changes requested):
In every test this same error:
TypeError: Cannot read properties of undefined (reading 'address')

danon321 (approved):

PR #511, opened 2024-01-30
solution

zaura333:


Patrykmclaren14 (approved):
All looks perfect!

mbruhler (changes requested):
Actually one thing - you can use [app.use](https://expressjs.com/en/4x/api.html#app.use) once before routes and then you wouldn't have to pass express.json() to every controller.

mbruhler (approved):
Well done!

PR #512, opened 2024-01-30
solution

bertsu02:


zaura333 (approved):

danon321 (changes requested):
Everything looks cool you used routes and services (thets great!), but what about controlers :/

Controlers should containr ur buisness logic so tell when response is 404/400 or 200
Services contain logic like filtering data

mbruhler (approved):
Good job, approved

PR #517, opened 2024-01-30
solution

Timi2424:


szymanski-marcin (approved):
Really nice code management. Wish I could do it like you:) Good job!

src/routes/expenses.routes.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const express = require('express');
+const cors = require('cors');
+
+const expensesController = require('../controllers/expenses.controller');
+
+const expensesRouter = express.Router();
+
+expensesRouter.use(cors());
```
mbruhler, line 10: This is not needed in this task, has no impact on the workflow, so I would delete that.
Also you do it just once in the createServer and that's enough

src/routes/users.routes.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const express = require('express');
+const cors = require('cors');
+const userController = require('../controllers/user.controller');
+
+const usersRouter = express.Router();
+
+usersRouter.use(cors());
```
mbruhler, line 9: Don't repet the middleware declarations

mbruhler: Also use approperiate and actually needed middlewares

mbruhler (changes requested):
That's cool that you use `app.use` rather than passing middlewares everytime to every controller.

nonetheless, you should also use only required for the bussiness case or technological problem middlewares, `cors` is not needed to accomplish this task. If somethig is not needed, a rule of thumb is to get rid of it.

So please delete uneeded cors middleware also their declaration

Also remember, that you have to declare the middleware only once, preferably in createServer function

mbruhler (approved):
Good job! Approved

PR #518, opened 2024-01-31
task

szymanski-marcin:


adrian200002 (approved):

Centmsn (changes requested):
While the code itself is working properly 👍 I strongly encourage you to split it into separate directories such as routes, controllers, and services.

Centmsn (approved):

PR #519, opened 2024-01-31
Solution.

adrian200002:


marcinzrodlowski (approved):
Looks great, tests are passing, nice job 💪

src/createServer.js
```diff
@@ -1,11 +1,20 @@
...
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  resetExpenses();
+  resetUsers();
+
+  app.use(express.json());
```
mbruhler, line 14: if you use this here, you don't have to pass it again to route

mbruhler (approved):
Overall good job,

I will approve, but remember that when you use `app.use` it's added globally as middleware and you don't have to pass it again to every route.

PR #520, opened 2024-01-31
task solution

thesidorenko:


src/user/user.router.js
```diff
@@ -0,0 +1,67 @@
...
+const express = require('express');
+
+const userService = require('./user.service.js');
+
+const userRouter = express.Router();
+
+userRouter.get('/', (request, response) => {
+  const users = userService.getUsers();
+
+  response.send(users);
```
DarkMistyRoom, line 13: move this logic to a controller. Same for all endpoints

src/expense/expense.service.js
```diff
@@ -0,0 +1,95 @@
...
+        new Date(expense.spentAt).valueOf() >= new Date(from).valueOf()
+    );
+  }
+
+  if (to) {
+    filteredExpenses = filteredExpenses.filter(
+      (expense) =>
+        new Date(expense.spentAt).valueOf() <= new Date(to).valueOf()
+    );
+  }
```
DarkMistyRoom, line 34: do we really need to filter expenses several times? Could we use `filter` only one time?

DarkMistyRoom (changes requested):
Good job! But let's improve the code a bit

src/user/user.router.js
thesidorenko: The router stands for the controller in this case. We do not need to add more files and split logic because it's already split enough.
We could also rename routes for controllers here but it will interfere with import naming.

What I was trying to say is that routes already handle all of the request/response operations and build as slim as it is possible.

src/expense/expense.service.js
thesidorenko: In this case, I used a filter several times to ensure that every search option could be used separately and combined in any configuration.
Also, it was the easiest way to accomplish such functionality.
I am aware that I have several loops and it uses a lot of memory therefore reducing performance. But we already store data in memory))

src/user/user.router.js
Andriy-Shymkiv: I agree with @DarkMistyRoom , because the point of this task is to understand how to build an MVC application, so please, create a controller for each entity and move this logic there as was mentioned above

src/expense/expense.service.js
Andriy-Shymkiv: Imagine that you have an array of one million records and the client trying to get them from api with filters like this one. It will filter such a big array 4 times, it is a lot of memory.

So you really need to iterate it only once

Andriy-Shymkiv (changes requested):

xiimiik (approved):
lgtm

PR #521, opened 2024-02-01
🌵

Masha-Cactus:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,79 @@
...
+  if (!expenseService.getById(id)) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const newExpense = expenseService.remove(id);
+
+  res.sendStatus(204);
+  res.send(newExpense);
```
Andriy-Shymkiv, line 70: the `remove` method returns undefined. so what did you trying to send to the client?

src/controllers/user.controller.js
```diff
@@ -0,0 +1,79 @@
...
+  if (!userService.getById(id)) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const newUsers = userService.remove(id);
+
+  res.sendStatus(204);
+  res.send(newUsers);
```
Andriy-Shymkiv, line 70: same as above

Andriy-Shymkiv (changes requested):
in a real application you will need to use some database with proper error handling, but for this application it's enough. good job! left a few comments

l4st1m0za (approved):
GJ!

PR #522, opened 2024-02-01
Solution

marcinzrodlowski:


src/services/expenses.js
```diff
@@ -0,0 +1,71 @@
...
+function getById(expensesId) {
+  const foundExpenses = expenses.find(
+    expense => expense.id === expensesId
+  );
+
+  return foundExpenses || null;
+}
+
+function create(userId, spentAt, title, amount, category, note) {
+  const newExpenses = {
```
kate-prakofyeva, line 33: const newExpense = {
    id: expenses.length + 1, // exp.length === last element in array + 1
    ...expense,
  };
  
  JUST FYI, you can write the same shorter

kate-prakofyeva (approved):

src/services/expenses.js
kate-prakofyeva: Good job!

src/controllers/expenses.js
```diff
@@ -0,0 +1,124 @@
+'use strict';
+
+const userService = require('../services/users.js');
+const expensesService = require('../services/expenses.js');
+
+const getAll = (req, res) => {
+  const { userId, categories, from, to } = req.query;
+
+  const expenses = expensesService.getAll({
```
mbruhler, line 9: you don't have to name every property when the whole payload is meant to be send, use just req.body here

src/controllers/expenses.js
```diff
@@ -0,0 +1,124 @@
...
+  const user = userService.getById(userId);
+
+  if (!user) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newExpenses = expensesService.create(
+    userId, spentAt, title, amount, category, note
```
mbruhler, line 47: same here

src/services/expenses.js
mbruhler: true

mbruhler (changes requested):
In overall, good solution but parts of your code are not much sustainable. Don't desctructure everywhere and name the property implicity, it's better to pass the whole request body to the another part of application rather than naming every property (some of the places pointed in comments, but check in overall). 

This way when somethig extra will be added, you won't have to type it everywhere again.

Also think of places in your code when you can make it more automatic

src/services/expenses.js
marcinzrodlowski: Sure but it's more readable for me

PR #524, opened 2024-02-03
Express task

VKdrvtsv:
have some problem with node moduls

### DIFF
--- a/src/controllers/users.controller.js
+++ b/src/controllers/users.controller.js
@@ -1,0 +1,112 @@
+'use strict';
+
+const {
+  getUsers,
+  createUser,
+  getUser,
+  deleteUser,
+  updateUser,
+} = require('../services/users.serveces');
+
+const getAllUsers = (req, res) => {
+  res.statusCode(200);
+  res.send(getUsers());
+};
+
+const createOneUser = (req, res) => {
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const todo = createUser(name);
+
+  res.statusCode(201);
+
+  res.send(todo);
+};
+
+const getOneUser = (req, res) => {
+  const { id } = req.params;
+
+  if (!id) {
+    res.sendStatus(400);
+  }
+
+  const user = getUser(id);
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.statusCode(200);
+  res.send(user);
+};
+
+const deleteOneUser = (req, res) => {
+  const { id } = req.params;
+
+  if (!id) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const user = getUser(id);
+
+  if (!user) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  deleteUser(id);
+
+  res.sendStatus(204);
+};
+
+const updateOneUser = (req, res) => {
+  const { id } = req.params;
+  const { name } = req.body;
+
+  if (!id) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const user = getUser(id);
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  if (typeof name !== 'string') {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const updatedUser = updateUser({
+    id,
+    name,
+  });
+
+  res.sendStatus(200);
+  res.send(updatedUser);
+};
+
+module.exports = {
+  getAllUsers,
+  createOneUser,
+  getOneUser,
+  deleteOneUser,
+  updateOneUser,
+};
--- a/src/services/users.serveces.js
+++ b/src/services/users.serveces.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const { v4 } = require('uuid');
+
+let users = [];
+
+const getUsers = () => users;
+
+const createUser = (name) => {
+  const user = {
+    name,
+    id: v4(),
+  };
+
+  users.push(user);
+
+  return user;
+};
+
+const getUser = (id) => {
+  return users.find(user => user.id === id) || null;
+};
+
+const deleteUser = (id) => {
+  const newUsers = users.filter(user => user.id !== id);
+
+  users = newUsers;
+};
+
+const updateUser = ({ id, name }) => {
+  const user = getUser(id);
+
+  Object.assign(user, { name });
+
+  return user;
+};
+
+module.exports = {
+  getUsers,
+  createUser,
+  getUser,
+  deleteUser,
+  updateUser,
+};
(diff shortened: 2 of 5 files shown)


src/services/users.serveces.js
```diff
@@ -0,0 +1,44 @@
...
+const deleteUser = (id) => {
+  const newUsers = users.filter(user => user.id !== id);
+
+  users = newUsers;
+};
+
+const updateUser = ({ id, name }) => {
+  const user = getUser(id);
+
+  Object.assign(user, { name });
```
l4st1m0za, line 33: The updateUser function directly modifies the user object. This might lead to unintended side effects. Consider creating a new object with the updated values.

src/controllers/users.controller.js
```diff
@@ -0,0 +1,112 @@
...
+  res.statusCode(201);
+
+  res.send(user);
+};
+
+const getOneUser = (req, res) => {
+  const { id } = req.params;
+
+  if (!id) {
+    res.sendStatus(400);
```
l4st1m0za, line 36: You should add return after `res.sendStatus(400)` to stop the execution of the function if id is not provided.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,153 @@
...
+
+  if (!id) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const expense = getExpense(id);
+
+  if (!expense) {
```
l4st1m0za, line 89: The status code should be 404 not 400 when an expense is not found.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,153 @@
...
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.statusCode(200);
```
l4st1m0za, line 74: Use the method status to set the status code, not statusCode.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,153 @@
...
+  res.send(expense);
+};
+
+const deleteOneExpense = (req, res) => {
+  const { id } = req.params;
+
+  if (!id) {
```
l4st1m0za, line 81: The status code should be 400 not 404 when id is not provided in the request.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,153 @@
...
+const getAllExpenses = (req, res) => {
+  const {
+    userId,
+    categories,
+    from,
+    to,
+  } = req.body;
+
+  if (!userId || !categories || !from || !to) {
+    res.sendStatus(400);
```
l4st1m0za, line 20: It's better to also provide a description of what went wrong, not just the status code.

l4st1m0za (changes requested):
Good job, but some fixes needed

src/services/users.serveces.js
VKdrvtsv: To be honest, I don't understand this comment because in this case we should to delete the real user from array and to create the new one? But if it will not lead to the same side effects? Now there aren`t mentors on the platform, so maybe you could help me)

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,154 @@
...
+    userId,
+    title,
+    spentAt,
+    category,
+    amount,
+    note,
+  } = req.body;
+
+  if (!userId || !category || !spentAt || !title || !amount) {
+    res.sendStatus(400);
```
vadiimvooo, line 41: Here you should also provide a description.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,154 @@
...
+  res.statusCode(201);
+
+  res.send(expense);
+};
+
+const getOneExpense = (req, res) => {
+  const { id } = req.params;
+
+  if (!id) {
+    res.sendStatus(400);
```
vadiimvooo, line 64: Here you should also provide a description.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,154 @@
...
+  const {
+    title,
+    spentAt,
+    category,
+    amount,
+    note,
+  } = req.body;
+
+  if (!id) {
+    res.sendStatus(400);
```
vadiimvooo, line 112: Here you should also provide a description.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,154 @@
...
+
+    return;
+  }
+
+  if (typeof title !== 'string'
+    || typeof category !== 'string'
+    || typeof note !== 'string'
+    || typeof spentAt !== 'string'
+    || typeof amount !== 'number') {
+    res.sendStatus(400);
```
vadiimvooo, line 130: Here you should also provide a description.

src/controllers/users.controller.js
```diff
@@ -0,0 +1,114 @@
...
+const getAllUsers = (req, res) => {
+  res.statusCode(200);
+  res.send(getUsers());
+};
+
+const createOneUser = (req, res) => {
+  const { name } = req.body;
+
+  if (!name) {
+    res.sendStatus(400);
```
vadiimvooo, line 20: Here you should also provide a description.

src/controllers/users.controller.js
```diff
@@ -0,0 +1,114 @@
...
+  const user = getUser(id);
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  if (typeof name !== 'string') {
+    res.sendStatus(400);
```
vadiimvooo, line 94: Here you should also provide a description.

src/controllers/users.controller.js
```diff
@@ -0,0 +1,114 @@
...
+
+  res.sendStatus(204);
+};
+
+const updateOneUser = (req, res) => {
+  const { id } = req.params;
+  const { name } = req.body;
+
+  if (!id) {
+    res.sendStatus(400);
```
vadiimvooo, line 80: Here you should also provide a description.

src/controllers/users.controller.js
```diff
@@ -0,0 +1,114 @@
...
+  if (!id) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  const user = getUser(id);
+
+  if (!user) {
+    res.sendStatus(400);
```
vadiimvooo, line 65: Here you should also provide a description.

src/services/users.serveces.js
```diff
@@ -0,0 +1,44 @@
...
+
+  return user;
+};
```
vadiimvooo, line 36: It can be done in such way. You can merge findIndex of user with getUser func as well.

```suggestion
const updateUser = ({ id, name }) => {
  const user = getUser(id);

  const updatedUser = { ...user, name };
  
  const indexOfUser = users.findIndex(user => user.id === id);
  
  users.splice(indexOfUser, 1, updatedUser);

  return updatedUser;
};
```

vadiimvooo (changes requested):
Well done. A few problems needs to be fixed.

vadiimvooo (approved):
Well done!)

PR #526, opened 2024-02-05
basic solution, in one file

jacekkosiba:


mbruhler (changes requested):
Good job, it's working, but your app should be organized into routes, controllers and services as it was told on lectures. In that shape I can't approve your PR.

mbruhler (approved):
Good job, approved

PR #530, opened 2024-02-11
add task solution

akozlovska:


src/utils/validateRequest.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const validateExpenseGetRequestQuery = ({ userId, categories, from, to }) => {
+  return (userId ? !isNaN(parseInt(userId)) : true)
+  && (categories
+    ? (Array.isArray(categories) || typeof categories === 'string')
+    : true)
+  && (from ? !isNaN(new Date(from)) : true)
+  && (to ? !isNaN(new Date(to)) : true);
```
DarkMistyRoom, line 9:
```suggestion
  return (userId ? !isNaN(parseInt(userId)) : true)
    && (categories
      ? (Array.isArray(categories) || typeof categories === 'string')
      : true)
    && (from ? !isNaN(new Date(from)) : true)
    && (to ? !isNaN(new Date(to)) : true);
```

src/controllers/user.controller.js
```diff
@@ -0,0 +1,93 @@
...
+  userService.remove(id);
+  res.sendStatus(204);
+};
+
+const updateOne = (req, res) => {
+  const id = parseInt(req.params.id);
+  const { name } = req.body;
+
+  if (isNaN(id) || typeof name !== 'string') {
+    res.sendStatus(400);
```
DarkMistyRoom, line 68: if `typeof name !== 'string'` better to send 422

src/service/expense.service.js
```diff
@@ -0,0 +1,83 @@
...
+
+  if (from) {
+    expensesFiltered = expensesFiltered
+      .filter(expense => new Date(expense.spentAt) >= new Date(from));
+  }
+
+  if (to) {
+    expensesFiltered = expensesFiltered
+      .filter(expense => new Date(expense.spentAt) <= new Date(to));
+  }
```
DarkMistyRoom, line 35: would be more efficient to apply only one filter with multiple conditions

DarkMistyRoom (changes requested):
Great job! A few comments:

src/controllers/user.controller.js
akozlovska: `typeof name !== 'string'` checks for correct data type as well as for name not being undefined (being provided by client) and in this case, according to task requirements I should send 400. Should I distinguish validation for existence and validation for correct data type and send 400 and 422 respectively in all responses?

Andriy-Shymkiv (approved):
very good!

PR #533, opened 2024-02-16
add task solution

PavloMykhalov:


src/services/expense.service.js
```diff
@@ -0,0 +1,82 @@
...
+
+  return newExpense;
+};
+
+const update = (id, data) => {
+  const expense = getById(id);
+
+  Object.assign(expense, { ...data });
+
+  return expense;
```
witflash, line 68: You can use the simpler syntax:
```
const update = (id, data) => ({
  ...getById(id),
  ...data
})
```
It's not a problem, just more readable for me.

src/services/expense.service.js
```diff
@@ -0,0 +1,82 @@
...
+
+      return spentAt >= fromDate && spentAt <= toDate;
+    });
+  }
+
+  return filteredExpenses;
+};
+
+const getById = (id) => {
+  return expenses.find(expense => expense.id === id);
```
witflash, line 37: You can use implicit return instead of directly writing `return`.

witflash (approved):
Good job!

Pay attention to my comments for your future work.

PR #540, opened 2024-03-11
node_accounting-app-v2

serjgrm:


src/controllers/expanses.js
```diff
@@ -0,0 +1,106 @@
...
+const { findUser } = require('../services');
+const { users } = require('./users');
+
+const expenses = [];
+
+function createExpense(req, res) {
+  const { userId, spentAt, title, amount, category, note } = req.body;
+
+  if (!userId || !spentAt || !title || !amount || !category
+    || !note || !findUser(userId, users)) {
```
Andriy-Shymkiv, line 12: are you sure that all params are required?

src/controllers/expanses.js
```diff
@@ -0,0 +1,106 @@
...
+      return spentAt >= fromDate && spentAt <= toDate;
+    });
+  }
+
+  if (categories) {
+    const selectedCategories = categories.split(',');
+
+    filteredExpenses = filteredExpenses
+      .filter(expense => selectedCategories.includes(expense.category));
+  }
```
Andriy-Shymkiv, line 51: you filtering multiple time here, please iterate only once

tests/expense.test.js
```diff
@@ -10,6 +11,7 @@ describe('Expense', () => {
   beforeEach(() => {
     server = createServer();
     api = supertest(server);
+    expenses.length = 0;
```
Andriy-Shymkiv, line 14: is it required?

Andriy-Shymkiv: I don't think that the tests is broken so please remove it

tests/user.test.js
```diff
@@ -10,6 +11,7 @@ describe('User', () => {
   beforeEach(() => {
     server = createServer();
     api = supertest(server);
+    users.length = 0;
```
Andriy-Shymkiv, line 14: same

Andriy-Shymkiv (changes requested):

tests/expense.test.js
l4st1m0za, line 14: Directly modifying the length of expenses array to 0 can lead to unintended behavior.

tests/user.test.js
l4st1m0za, line 14: Directly setting the length of the users array to 0 may not be a good way to ensure a clean test environment.

l4st1m0za (approved):
Good job

PR #543, opened 2024-03-12
add solution

nabokok:


dyarechuk (approved):
Impressive work, approving)🔥
PrushkaMadeInAbyssGIF

src/services/user.service.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+let users = [];
```

src/services/expenses.service.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+let expenses = [];
```
DanilWeda, line 3: Also about this place. 
Try use const.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,88 @@
...
+  const { userId, categories, from, to } = req.query;
+
+  res.send(expensesService.getExpenses(userId, categories, from, to));
+};
+
+const getOne = (req, res) => {
+  const { id } = req.params;
+
+  if (!id) {
+    res.sendStatus(400);
```
DanilWeda, line 16: Use constants for all status codes.

DanilWeda (changes requested):
Good job! 
Check comments for improvements. 
Thanks.

src/services/user.service.js
nabokok: I reassign users in deleteUser func

src/services/expenses.service.js
nabokok: Also reassigns

DanilWeda (approved):
Approved.
Thanks.

PR #544, opened 2024-03-12
Add task solution

dyarechuk:


src/services/expense.service.js
```diff
@@ -0,0 +1,78 @@
...
+const { getId } = require('./getId');
+
+const expenses = [];
+
+const init = () => {
+  expenses.length = 0;
+};
+
+const getAll = (params) => {
+  let filteredExpenses = expenses;
```
rvoshchylo, line 12: better use const here (optional) (not critical)

rvoshchylo (approved):
Not bad, ma frendo, but u can do it better!

src/services/expense.service.js
DanilWeda: Agreed. 
Why do we need let in this case? 

PR #546, opened 2024-03-12
add task solution

rvoshchylo:


sasholk (approved):
oiboi u are so good! 💦💦💦

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,106 @@
...
+};
+
+const deleteExpenseController = (req, res) => {
+  const { id } = req.params;
+
+  const checkedExpense = getOneExpense(id);
+  const expenseUpdateIndex = findIndexOneExpense(checkedExpense);
+
+  if (expenseUpdateIndex < 0) {
+    res.sendStatus(404);
```
DanilWeda, line 92: Tip* -  you can create object ( in ts - enum) - STATUS_CODES or something like that)

DanilWeda (changes requested):
rc*

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,111 @@
+'use strict';
+
+const NOT_FOUND = 404;
+const BAD_REQUEST = 400;
+const CREATED = 201;
+const NO_CONTENT = 204;
```
DanilWeda, line 6: You has duplicate in this place

src/controllers/user.controller.js
```diff
@@ -0,0 +1,100 @@
+'use strict';
+
+const NOT_FOUND = 404;
+const BAD_REQUEST = 400;
+const CREATED = 201;
+const NO_CONTENT = 204;
+const OK = 200;
```
DanilWeda, line 7: And this place. 
Please add it to the variable in the global scope, and use it in all places. 
Thanks)

DanilWeda (approved):
Great! 
Approved.

PR #547, opened 2024-03-12
add solution

pivkopa:


AntonDuchenko (approved):
Great job!
Approved!
But I think will be better to separate createServer.js into diff folders and files, like router, controllers and services!

src/createServer.js
```diff
@@ -1,11 +1,207 @@
+/* eslint-disable no-console */
 'use strict';
 
-// const express = require('express');
+const express = require('express');
 
 function createServer() {
```
olya-shyian, line 6: please split the code into smaller peaces according to Anton's advice

src/createServer.js
```diff
@@ -1,11 +1,207 @@
...
+  const findUserById = (userId) => users.find(
+    user => user.id === userId);
+  const findExpenseById = (expenseId) => expenses.find(
+    expense => expense.id === expenseId);
+
+  app.post('/expenses', (req, res) => {
+    const { userId, spentAt, title, amount, category, note } = req.body;
+
+    if (!userId || !spentAt || !title || !amount || !category || !note) {
+      res.status(400).send({ message: 'Required parameter(s) missing' });
```
olya-shyian, line 23: create consts for code statuses

package.json
```diff
@@ -20,11 +20,13 @@
...
-    "mocha": "^10.1.0",
+    "mocha": "^10.3.0",
     "nodemon": "^2.0.20",
     "supertest": "^6.3.0"
   },
   "dependencies": {
-    "express": "^4.18.2"
+    "cors": "^2.8.5",
+    "express": "^4.18.3",
+    "uuid": "^9.0.1"
```
olya-shyian, line 30: I see that you installed `uuid`, then maybe better to use it or uninstall if you changed your mind

olya-shyian (changes requested):
Fantastic! Well done!💥

src/constrollers/expenses.controller.js
```diff
@@ -0,0 +1,118 @@
...
+  );
+
+  res.status(OK).send(updatedExpense);
+};
+
+const remove = (req, res) => {
+  const expenseId = parseInt(req.params.id);
+
+  if (!expenseService.getById(expenseId)) {
+    return res.status(NOT_FOUND).send({ message: 'Expense not found' });
```
DanilWeda, line 105: You can add all message in constant

src/services/user.service.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+let users = [];
```

src/services/expenses.service.js
```diff
@@ -0,0 +1,73 @@
+/* eslint-disable no-console */
+'use strict';
+
+let expenses = [];
```
DanilWeda, line 4: Why do we need let in this case?

DanilWeda (changes requested):
Good job. 
Check improvements. 
Thanks)

DanilWeda (approved):
Great! 
Approved. 
Thanks)

PR #548, opened 2024-03-12
solution

kostrubitskii:


yaneznayou (approved):
good job 
interesting solution
i can't see any problems here
approved

src/services/user.service.js
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable no-console */
+'use strict';
+
+let users = [];
```
DanilWeda, line 4: Why do we need let in this case?

( I see reassign in code, but in this case you also can simply use const)

src/services/expense.service.js
```diff
@@ -0,0 +1,84 @@
+/* eslint-disable no-console */
+'use strict';
+
+let expenses = [];
```
DanilWeda, line 4: Why do we need let in this case?

( I see reassign in code, but in this case you also can simply use const)

src/routes/expense.route.js
```diff
@@ -0,0 +1,101 @@
...
+  res.send(expense);
+});
+
+expensesRouter.post('/', (req, res) => {
+  const { userId, spentAt, title, amount, category, note, id } = req.body;
+
+  const user = getById(userId);
+
+  if (!user || !title) {
+    res.sendStatus(400);
```
DanilWeda, line 27: Add all status code in constants

DanilWeda (approved):
Great! 
Approved.

PR #549, opened 2024-03-13
add task solution

yaneznayou:
add task solution

### DIFF
--- a/src/expense/expenses.controller.js
+++ b/src/expense/expenses.controller.js
@@ -1,0 +1,106 @@
+'use strict';
+
+const {
+  getAllExpenses,
+  getOneExpense,
+  checkAtleastOneUser,
+  createExpense,
+  findIndexOneExpense,
+  deleteExpense,
+  updateExpense,
+} = require('../services/expenses.service');
+
+const getAll = (req, res) => {
+  const { userId, categories, from, to } = req.query;
+
+  const expensesQuery = getAllExpenses(userId, categories, from, to);
+
+  res.send(expensesQuery);
+};
+
+const getOneById = (req, res) => {
+  const { id } = req.params;
+
+  const expenseIdFound = getOneExpense(id);
+
+  if (!expenseIdFound) {
+    res.status(404).send('Expense not found');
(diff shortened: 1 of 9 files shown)


src/expense/expenses.controller.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  res.send(expenseIdFound);
+};
+
+const createExpenseController = (req, res) => {
+  const { title, category, note, amount, userId, spentAt } = req.body;
+
+  if (!title || !category || !note || !amount || !userId || !spentAt) {
+    res.status(400).send('Invalid request body');
```
OleksandrRezanov, line 37: note is optional. You don't need to check it

src/expense/expenses.controller.js
```diff
@@ -0,0 +1,106 @@
...
+  res.send(expensesQuery);
+};
+
+const getOneById = (req, res) => {
+  const { id } = req.params;
+
+  const expenseIdFound = getOneExpense(id);
+
+  if (!expenseIdFound) {
+    res.status(404).send('Expense not found');
```
OleksandrRezanov, line 27: It might be better to add all status codes and messages to constants.

OleksandrRezanov (approved):
Good job! Keep it up!

src/expense/expenses.controller.js
DanilWeda: Agreed. Create constants for all status codes.

src/user/user.controller.js
```diff
@@ -0,0 +1,94 @@
...
+  res.status(200).send(userCheck);
+};
+
+const deleteUserId = (req, res) => {
+  const { userId } = req.params;
+
+  const userToDelete = getUserById(userId);
+
+  if (!userToDelete) {
+    res.status(404).send('User not found');
```
DanilWeda, line 76: Move all errors to constants. Thanks)

DanilWeda (changes requested):
Good job. 
Check some improvements. 
Thanks)

PR #550, opened 2024-03-13
solution att#1.

Manankin:
But not all tests done; continue to fix them

### DIFF
--- a/src/controllers/expense.controller.js
+++ b/src/controllers/expense.controller.js
@@ -48,7 +48,7 @@ function get(req, res) {
 
 function getOneExpense(req, res) {
   const { id } = req.params;
-  const expense = getExpenseById(id);
+  const expense = getExpenseById(+id);
 
   if (!expense) {
     res.sendStatus(404);
@@ -69,7 +69,13 @@ function addExpense(req, res) {
     note,
   } = req.body;
 
-  const user = getAllUsers.find(u => u.id === +userId);
+  if (!title || !category || !note || !amount || !userId || !spentAt) {
+    res.status(400).send('Invalid request body');
+
+    return;
+  }
+
+  const user = getAllUsers().find(u => u.id === +userId);
 
   if (!title || !user) {
     res.sendStatus(400);
@@ -86,16 +92,17 @@ function addExpense(req, res) {
     note,
   });
 
-  res.status(201).json(newExpense);
+  console.log(newExpense);
+  res.status(201).send(newExpense);
 }
 
 function deleteExpense(req, res) {
   const { id } = req.params;
 
-  const newExpenses = deleteExpenseById(id);
+  const newExpenses = deleteExpenseById(+id);
 
   if (newExpenses.length === getAllExpenses().length) {
-    res.status(404);
+    res.sendStatus(404);
 
     return;
   }
@@ -106,8 +113,8 @@ function deleteExpense(req, res) {
 
 function changeExpense(req, res) {
   const { id } = req.params;
-  const { title } = req.body;
-  const expense = getExpenseById(id);
+  const { ...updatedParams } = req.body;
+  const expense = getExpenseById(+id);
 
   if (!expense) {
     res.sendStatus(404);
@@ -115,7 +122,7 @@ function changeExpense(req, res) {
     return;
   }
 
-  const updatedExpense = updateExpense(expense, title);
+  const updatedExpense = updateExpense(expense, updatedParams);
 
   return res.send(updatedExpense);
 }
--- a/src/controllers/expense.controller.js
+++ b/src/controllers/expense.controller.js
@@ -1,0 +1,129 @@
+'use strict';
+
+const { getAllUsers } = require('../services/user.service');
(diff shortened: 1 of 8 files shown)


YevhenOstrovskiy (changes requested):
Looks nice, but not all test have passed!!!
Check how you get id, because you use array length, but if we delete some item and then add another one we can get the same id's. Try to use something like this:
 id: users.length > 0
      ? +(users[users.length - 1].id) + 1 : users.length + 1
      or use 'uuid' library from express instead

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,136 @@
...
+
+  if (!title || !user) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const newExpense = createExpense({
+    userId,
+    spentAt,
```
DanilWeda, line 88: Please fix this lint error or disabled rule

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,136 @@
...
+    note,
+  } = req.body;
+
+  if (!title || !category || !note || !amount || !userId || !spentAt) {
+    res.status(400).send('Invalid request body');
+
+    return;
+  }
+
+  const user = getAllUsers().find(u => u.id === +userId);
```
DanilWeda, line 78: Fix naming. u - blurry meaning

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,136 @@
...
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
+  } = req.body;
+
+  if (!title || !category || !note || !amount || !userId || !spentAt) {
+    res.status(400).send('Invalid request body');
```
DanilWeda, line 73: Add all status codes and message in constants

src/services/user.service.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+let users = [];
+let increment = 1;
```
DanilWeda, line 4: Why do we need let in this case?

src/services/expense.service.js
```diff
@@ -0,0 +1,64 @@
+'use strict';
+
+let expenses = [];
+let increment = 1;
```
DanilWeda, line 4: Why do we need let in this case?

src/controllers/user.controller.js
```diff
@@ -0,0 +1,88 @@
...
+  const { name } = req.body;
+  const user = getUserById(+id);
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  if (typeof name !== 'string') {
```
DanilWeda, line 71: Try not to use type for condition. 
Don`t see any reason for this type of condition. 
Maybe I'm missing something?

src/services/expense.service.js
```diff
@@ -0,0 +1,64 @@
...
+function clearAllExpenses() {
+  expenses = [];
+}
+
+function getAllExpenses() {
+  return expenses;
+}
+
+function getExpenseById(id) {
+  return expenses.find(exp => exp.id === id);
```
DanilWeda, line 15: fix naming

src/services/expense.service.js
```diff
@@ -0,0 +1,64 @@
...
+  };
+
+  expenses.push(newExpense);
+  increment++;
+
+  return newExpense;
+}
+
+function deleteExpenseById(id) {
+  const filteredExpenses = expenses.filter(exp => exp.id !== id);
```
DanilWeda, line 43: fix naming

DanilWeda (changes requested):
Good job! 
Check some small issues) 
Also, check test suite.
Many thanks!

src/controllers/expense.controller.js
Manankin: it's naming inside callback function and I don't use it outside.  guess that it's clear to understand (user => u)

src/services/expense.service.js
Manankin: the same reason as above

src/services/user.service.js
DanilWeda: I see this assignment. 
We can manipulate with an array. 
Any let assigned makes code more dirty readability.
Please try rewrite, also in this case you don't need ` let increment = 1;` part of code.
Thanks!

src/services/expense.service.js
DanilWeda: I see this assignment. 
We can manipulate with an array. 
Any let assigned makes code more dirty readability.
Please try rewrite, also in this case you don't need let increment = 1; part of code.
Thanks!

DanilWeda (changes requested):
Check my answer to your comments. 
Thanks)

DanilWeda (changes requested):
Great. All good. 
One small thing for better understanding code. 
Many thanks)

DanilWeda (approved):
Great! 
Approved. 
Thanks)

PR #551, opened 2024-03-13
solution

avakiel:


l0ad1nggg (approved):
well done!

src/Data/Store.js
```diff
@@ -0,0 +1,117 @@
...
+    const expenseIndex = this.expenses.findIndex(expense => expense.id === +id);
+
+    this.expenses[expenseIndex] = {
+      ...this.expenses[expenseIndex],
+      spentAt: spentAt || this.expenses[expenseIndex].spentAt,
+      title: title || this.expenses[expenseIndex].title,
+      amount: amount || this.expenses[expenseIndex].amount,
+      category: category || this.expenses[expenseIndex].category,
+      note: note || this.expenses[expenseIndex].note,
+    };
```
DanilWeda, line 106: Has some questions about this functionality. 
Try rewriting it, using the spread/rest operator.
Prepared object with value, then spread this object using an index. 
Why do we need this index spread `...this.expenses[expenseIndex]`? 
We currently find element and work with them, and their value. I think a spread operator isn't needed. 
Also for safety updates, you need to work only with information that the user gives you. Try to create an object with user-correct input.   And then spread operator change only these values, the rest will be not updated.

src/Routers/ExpensesRouter.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  res.status(200).send(myStore.getExpenses(id));
+});
+
+expensesRouter.delete('/:id', (req, res) => {
+  const { id } = req.params;
+
+  if (myStore.getExpenses(id) === undefined) {
+    res.sendStatus(404);
```
DanilWeda, line 49: Create constants for all status code

src/Routers/UsersRouter.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  res.status(200).send(myStore.getUser(userId));
+});
+
+usersRouter.delete('/:userId', (req, res) => {
+  const { userId } = req.params;
+  const deleteUser = myStore.getUser(userId);
+
+  if (deleteUser === undefined) {
```
DanilWeda, line 31: Try not to use type condition.

DanilWeda (changes requested):
Good job. 
Check some comments for improvements. 
Thanks)

src/Services/User.services.js
```diff
@@ -0,0 +1,51 @@
+'use strict';
+
+const { generateUniqueID } = require('../utils/IdGenerator');
+
+let users = [];
```
DanilWeda, line 5: Why do we need let in this case?

src/Services/Expenses.services.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+const { generateUniqueID } = require('../utils/IdGenerator');
+
+let expenses = [];
```
DanilWeda, line 5: Why do we need let in this case?

DanilWeda (changes requested):
All good! 
I liked your solution. 
One small thing. 
You use `let` when you operate with the array. 
I think you can simplify the solution, and use const instead, without using assigns) 
Thanks)

src/Services/User.services.js
avakiel: it was an easy way

DanilWeda (approved):
Great. 
Approved. 
Thanks)

PR #552, opened 2024-03-13
added solution

OleksandrRezanov:


src/services/expense.service.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+const expenses = [];
+
+const getMaxId = () => {
+  return expenses.length
+    ? Math.max(...expenses.map(expense => expense.id))
+    : 0;
+};
```
DanilWeda, line 9: See duplicate this

src/services/user.service.js
```diff
@@ -0,0 +1,51 @@
...
+  const index = users.findIndex(user => user.id === +id);
+
+  users.splice(index, 1);
+};
+
+const getMaxId = () => {
+  return users.length
+    ? Math.max(...users.map(user => user.id))
+    : 0;
+};
```
DanilWeda, line 42: and this place. 
Maybe create one helper for it, and use it in two places?

DanilWeda (changes requested):
Good job! 
Check small optional comment. 
Thanks)

DanilWeda (approved):
Great! 
Approved)

PR #553, opened 2024-03-13
add solution

l0ad1nggg:


src/services/users.service.js
```diff
@@ -0,0 +1,50 @@
...
+const updateUser = (id, newName) => {
+  const user = getUserById(id);
+
+  user.name = newName;
+
+  return user;
+};
+
+const removeUser = (id) => {
+  users = users.filter(u => u.id !== +id);
```
DanilWeda, line 36: fix naming u - blurry meaning

src/services/users.service.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+let users = [];
```
DanilWeda, line 3: Why do we need let in this case?

src/services/expenses.service.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+
+let expenses = [];
```
DanilWeda, line 3: Why do we need let in this case?

src/controllers/user.controller.js
```diff
@@ -0,0 +1,69 @@
...
+  const user = userService.getUserById(id);
+
+  if (!user) {
+    res.status(404).send();
+
+    return;
+  }
+
+  userService.removeUser(+id);
+  res.status(204).send();
```
DanilWeda, line 60: add all status in constants

src/services/users.service.js
```diff
@@ -0,0 +1,50 @@
...
+
+const getAllUsers = () => {
+  return users;
+};
+
+const getUserById = (id) => {
+  const user = users.find(u => u.id === +id);
```
DanilWeda, line 10: fix naming

src/services/expenses.service.js
```diff
@@ -0,0 +1,96 @@
...
+
+const getAllExpenses = (userId, categories, from, to) => {
+  let filteredExpenses = [...expenses];
+
+  if (userId) {
+    filteredExpenses = filteredExpenses.filter(e => e.userId === +userId);
```
DanilWeda, line 9: fix naming

src/services/expenses.service.js
```diff
@@ -0,0 +1,96 @@
...
+      expense => new Date(expense.spentAt) >= new Date(from)
+        && new Date(expense.spentAt) <= new Date(to)
+    );
+  }
+
+  return filteredExpenses;
+};
+
+const getExpenseById = (id) => {
+  const expense = expenses.find(e => e.id === +id);
```
DanilWeda, line 28: fix naming

src/services/expenses.service.js
```diff
@@ -0,0 +1,96 @@
...
+
+  return expense;
+};
+
+const getExpenseByUserId = (userId) => {
+  const expense = expenses.filter(e => e.userId === +userId);
```
DanilWeda, line 34: fix naming

src/services/expenses.service.js
```diff
@@ -0,0 +1,96 @@
...
+  const filteredExpenses = expenses.filter(expense => {
+    return expense.spentAt >= from && expense.spentAt <= to;
+  });
+
+  return filteredExpenses;
+};
+
+const getExpenseByCategory = (userId, categories) => {
+  const expense = expenses
+    .filter(e => e.userId === userId && e.category === categories);
```
DanilWeda, line 49: fix naming

DanilWeda (changes requested):
Good job. 
Check some improvements. 
Thanks)

src/services/users.service.js
l0ad1nggg: I used let instead of const for deleting users and expenses from array, ok, I'll change to const

DanilWeda (approved):
Good! 
You skipped some comments, but I approved. 
but please, be attentive. 
Thanks)

PR #554, opened 2024-03-14
Develop

vpashko:


vSainiuk (approved):
Good job

src/services/expensesService.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+let expenses = [];
```
DanilWeda, line 3: You can use const in this place. 
I see you manipulate with an array, easy way to use const and avoid reassigned.

src/services/userService.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+let users = [];
```
DanilWeda, line 3: You can use const in this place. 
I see you manipulate with an array, easy way to use const and avoid reassigned.

src/services/userService.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+const getUserById = (id) => {
+  const userById = users.find(user => user.id === +id);
+
+  return userById;
+};
+
+const createUser = (name) => {
+  const id = Date.now();
```
DanilWeda, line 20: You can write a function to generate id, and use it and all places where you need it.

DanilWeda (changes requested):
Good job. 
Please check some small comments. 
Many thanks)

PR #555, opened 2024-03-14
Develop

YevhenOstrovskiy:


src/services/user.service.js
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  const changedUser = users.find(u => u.id === +id);
+
+  return changedUser;
+};
+
+const deleteUser = (id) => {
+  if (users.some(user => user.id === +id)) {
+    users = users.filter(u => u.id !== +id);
```
DanilWeda, line 49: fix naming, u a very blurry meaning

src/services/user.service.js
```diff
@@ -0,0 +1,60 @@
...
+    if (user.id === +id) {
+      user.name = name;
+
+      return user;
+    }
+
+    return user;
+  });
+
+  const changedUser = users.find(u => u.id === +id);
```
DanilWeda, line 42: fix naming, u a very blurry meaning

src/services/user.service.js
```diff
@@ -0,0 +1,60 @@
...
+const getUserById = (id) => {
+  const user = users.find(u => u.id === +id);
+
+  return user;
+};
+
+const createUser = (name) => {
+  const user = {
+    id: users.length > 0
+      ? +(users[users.length - 1].id) + 1 : users.length + 1,
```
DanilWeda, line 22: create func for generation id and use in 2 places were you need id

src/services/user.service.js
```diff
@@ -0,0 +1,60 @@
+'use strict';
+
+let users = [];
```

src/services/expenses.service.js
```diff
@@ -0,0 +1,93 @@
...
+};
+
+const createExpense = (newExpense) => {
+  const { userId, spentAt, title, amount, category, note } = newExpense;
+
+  const expens = {
+    id:
+      expenses.length > 0
+        ? +expenses[expenses.length - 1].id + 1
+        : expenses.length + 1,
```
DanilWeda, line 49: use func for generation in this place

src/services/expenses.service.js
```diff
@@ -0,0 +1,93 @@
...
+        ...updateExpense,
+      };
+
+      return expens;
+    }
+
+    return expens;
+  });
+
+  const changedExpens = expenses.find((e) => e.id === +id);
```
DanilWeda, line 77: fix naming

src/services/expenses.service.js
```diff
@@ -0,0 +1,93 @@
...
+
+  return changedExpens;
+};
+
+const deleteExpense = (id) => {
+  expenses = expenses.filter((e) => e.id !== +id);
```
DanilWeda, line 83: fix naming

src/services/expenses.service.js
```diff
@@ -0,0 +1,93 @@
+/* eslint-disable no-param-reassign */
+'use strict';
+
+let expenses = [];
```
DanilWeda, line 4: Why do we need let in this case?

src/controllers/user.controller.js
```diff
@@ -0,0 +1,71 @@
...
+};
+
+const remove = (req, res) => {
+  const { id } = req.params || undefined;
+
+  if (userService.getUserById(id)) {
+    userService.deleteUser(id);
+    res.status(statusCode.NO_CONTENT).send();
+  } else {
+    res.status(statusCode.NOT_FOUND).send(`User with id:${id} does not exist`);
```
DanilWeda, line 61: You can create constants for all your messages)

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,88 @@
...
+  res.send(expens);
+};
+
+const update = (req, res) => {
+  const { id } = req.params || undefined;
+  const updateExpense = req.body || undefined;
+
+  if (!expensesService.getExpensesById(id)) {
+    res.status(statusCode.NOT_FOUND);
+    res.send('Expense not found');
```
DanilWeda, line 60: For all files* ( about messages)

DanilWeda (changes requested):
Great! 
Check some improvements. 
Thanks)

DanilWeda (approved):
Great! 
Approved.
Thanks)

PR #559, opened 2024-03-18
add solution

MaxKalalashnyk02:


avakiel (approved):
Good job!

src/services/user.service.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+let users = [];
```
DanilWeda, line 3: You can use const in this place. 
I see you manipulate with an array, easy way to use const and avoid reassigned.

src/services/expenses.service.js
```diff
@@ -0,0 +1,69 @@
...
+
+  if (userId) {
+    filteredExpenses = filteredExpenses.filter(
+      expense => expense.userId === +userId
+    );
+  }
+
+  if (categories) {
+    filteredExpenses = filteredExpenses.filter(
+      expense => categories.includes(expense.category)
```
DanilWeda, line 20: Fixed or disabled linting

src/services/expenses.service.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+let expenses = [];
```
DanilWeda, line 3: You can use const in this place. 
I see you manipulate with an array, easy way to use const and avoid reassigned.

DanilWeda (changes requested):
Good job. 
Check small comments about using let and linter. 
Thanks)

PR #560, opened 2024-03-18
solution

levchenko-dmytro:


src/services/expense.service.js
```diff
@@ -0,0 +1,81 @@
...
+  if (from && to) {
+    const fromDate = new Date(from);
+    const toDate = new Date(to);
+
+    userExpenses = userExpenses.filter((exp) => {
+      return (
+        new Date(exp.spentAt) >= fromDate && new Date(exp.spentAt) <= toDate
+      );
+    });
+  }
```
vadiimvooo, line 33: It's better to filter the data once instead of doing it three times.

vadiimvooo (changes requested):
Almost done. Just one fix, please feel free to ask for help in fe_chat if you have any questions.

lerastarynets (approved):
Keep up the great work!

PR #563, opened 2024-03-27
Develop

boytrend0108:


src/controllers/user.controller.js
```diff
@@ -0,0 +1,72 @@
+const userServices = require('../services/user.service.js');
+
+const getUsers = (req, res) => {
+  res.send(userServices.getAll());
+};
+
+const getOne = (req, res) => {
+  const { id } = req.params;
```
DarkMistyRoom, line 8: let's send 400 if `id` isn't passed

DarkMistyRoom: the same for expenses

src/services/user.service.js
```diff
@@ -0,0 +1,61 @@
+let users = [
+  // { id: 2, name: 'Alice' },
+  // { id: 1, name: 'Bob' },
```
DarkMistyRoom, line 3: remove the comments

tests/expense.test.js
```diff
@@ -176,7 +176,7 @@ describe('Expense', () => {
       const response = await api
         // eslint-disable-next-line max-len
         .get(
-          `/expenses?&from=2022-10-19T00:00:00.000Z&to=2022-10-19T23:59:59.999Z`,
+          `/expenses?userId=${expenseData.userId}&from=2022-10-19T00:00:00.000Z&to=2022-10-19T23:59:59.999Z`,
```
DarkMistyRoom, line 179: Don't change tests. Initialize an empty arrays on each server start instead

src/services/user.service.js
```diff
@@ -0,0 +1,61 @@
...
+
+const getOne = (id) => {
+  const user = users.find((item) => item.id === +id) || null;
+
+  return user;
+};
+
+const create = (name) => {
+  if (typeof name !== 'string' || !name) {
+    users = [];
```
DarkMistyRoom, line 18: do we need to clear the whole array in this case?

src/services/expense.servise.js
```diff
@@ -0,0 +1,181 @@
...
+  // },
+  // {
+  //   userId: 1,
+  //   id: 40,
+  //   spentAt: '2022-11-21T11:01:43.462Z',
+  //   title: 'Buy a new laptop',
+  //   amount: 999,
+  //   category: 'Cars',
+  //   note: 'I need a new laptop',
+  // },
```
DarkMistyRoom, line 39: remove the comments

src/services/expense.servise.js
```diff
@@ -0,0 +1,181 @@
...
+    filtered = filtered.filter((el) => el.userId === +userId);
+  }
+
+  if (categories) {
+    filtered = filtered.filter((el) => categories.includes(el.category));
+  }
+
+  if (from) {
+    filtered = filtered.filter((el) => filterByPeriod(el, from, to));
+  }
```
DarkMistyRoom, line 55: can we reduce the amount of passes through the array?

DarkMistyRoom (changes requested):
Good job! Let's improve it a bit

vadiimvooo (approved):
LGTM

PR #565, opened 2024-04-01
solution

yuron-maker:


src/services/expenses.js
```diff
@@ -0,0 +1,76 @@
...
+  if (from && to) {
+    const dateFrom = new Date(from);
+    const dateTo = new Date(to);
+
+    filteredExpenses = filteredExpenses.filter(
+      (expense) =>
+        dateFrom <= new Date(expense.spentAt) &&
+        new Date(expense.spentAt) <= dateTo,
+    );
+  }
```
andriy-shymkiv, line 30: you use filter at least three times here. would be good to combine it to iterate only once

andriy-shymkiv (changes requested):

src/controllers/userController.js
```diff
@@ -0,0 +1,68 @@
+const usersService = require('../services/user');
+
+const get = (req, res) => {
+  res.statusCode = 200;
+  res.send(usersService.getAll());
+};
+
+const getById = (req, res) => {
+  const { id } = req.params;
+  const User = usersService.getById(id);
```
DarkMistyRoom, line 10:
```suggestion
  const user = usersService.getById(id);
```

DarkMistyRoom: we use camelCase for variables

src/controllers/userController.js
```diff
@@ -0,0 +1,68 @@
+const usersService = require('../services/user');
+
+const get = (req, res) => {
+  res.statusCode = 200;
+  res.send(usersService.getAll());
+};
+
+const getById = (req, res) => {
+  const { id } = req.params;
```
DarkMistyRoom, line 9: check if `id` is passed

src/controllers/userController.js
```diff
@@ -0,0 +1,68 @@
...
+  usersService.remove(id);
+
+  res.sendStatus(204);
+};
+
+const update = (req, res) => {
+  const { id } = req.params;
+  const { name } = req.body;
+
+  const User = usersService.update(id, name);
```
DarkMistyRoom, line 52:
```suggestion
  const user = usersService.update(id, name);
```

src/controllers/userController.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  usersService.remove(id);
+
+  res.sendStatus(204);
+};
+
+const update = (req, res) => {
+  const { id } = req.params;
+  const { name } = req.body;
```
DarkMistyRoom, line 50: check if these values were passed

src/controllers/expensesController.js
```diff
@@ -0,0 +1,86 @@
...
+
+  const Expense = expensesService.update(id, body);
+
+  if (!Expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+  res.send(Expense);
+};
```
DarkMistyRoom, line 78: the same suggestions as for userController

src/services/user.js
```diff
@@ -0,0 +1,56 @@
...
+  },
+  {
+    id: 'c61b9b1c-53ad-439d-8b9a-57eded061d1c',
+    name: 'proba2',
+  },
+  {
+    id: 'a34f5579-2917-429f-9f3b-313888fdf5b3',
+    name: 'proba3',
+  },
+];
```
DarkMistyRoom, line 15: remove the test data

src/createServer.js
```diff
@@ -1,11 +1,20 @@
...
-  // Return the server (express app)
+  const app = express();
+
+  app.use(express.json());
+  app.use(cors());
+
+  app.use('/users', usersRouter);
+  app.use('/expenses', expensesRouter);
+
+  return app;
```
DarkMistyRoom, line 17: initialize the users and expenses with empty arrays on each server start

DarkMistyRoom (changes requested):
Good job! Here are some cases for improvement:

src/controllers/expensesController.js
```diff
@@ -0,0 +1,105 @@
...
+  const { id } = req.params;
+  const body = req.body;
+
+  if (!id || !body) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  const Expense = expensesService.update(id, body);
```
DarkMistyRoom, line 89:
```suggestion
  const expense = expensesService.update(id, body);
```

src/controllers/expensesController.js
```diff
@@ -0,0 +1,105 @@
...
+  expensesService.remove(id);
+
+  res.sendStatus(204);
+};
+
+const update = (req, res) => {
+  const { id } = req.params;
+  const body = req.body;
+
+  if (!id || !body) {
```
DarkMistyRoom, line 83: `body` is an object so it will be always true. No need to check it here

DarkMistyRoom (approved):
Good job!

PR #569, opened 2024-04-19
add server user and expense api

AntonBelia:


src/controllers/user.controller.js
```diff
@@ -0,0 +1,69 @@
...
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.status(200).json(user);
+};
+
+const create = (req, res) => {
+  const body = req.body;
```
l4st1m0za, line 23: You can use object destructuring to access the properties of req.body. This helps make the code cleaner and easier to understand. For example: const { name } = req.body instead of const body = req.body

l4st1m0za (approved):
Good job!

PR #575, opened 2024-05-10
Node accounting app v2

nk8bnj:


src/services/expenses.service.js
```diff
@@ -0,0 +1,50 @@
+const { getFilteredExpenses } = require('../utils/getFilteredExpenses');
+
+let expenses = [];
+
+const expensesInit = () => {
+  expenses = [];
+};
```
BudnikOleksii, line 7: It looks unnecessary

src/services/expenses.service.js
```diff
@@ -0,0 +1,50 @@
...
+
+const getExpenses = (query) => {
+  expenses = getFilteredExpenses(expenses, query);
```
BudnikOleksii, line 10: You can't mutate data from your database/storage in get services.

src/utils/getFilteredExpenses.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    if (query.from && query.to) {
+      return (
+        new Date(expense.spentAt) > new Date(query.from) &&
+        new Date(expense.spentAt) < new Date(query.to)
+      );
+    }
+
+    return true;
```
BudnikOleksii, line 18: What if you have a few filters at the same time?

BudnikOleksii (changes requested):
Good job 👍 
Just a few moments should be fixed, you can't mutate the data when you need to just get it.

src/services/expenses.service.js
nk8bnj: It needs for update initial state.

src/routes/expenses.route.js
```diff
@@ -0,0 +1,15 @@
+const express = require('express');
+const expensesRouter = express.Router();
+const expensesController = require('../controllers/expenses.controller');
+
+expensesRouter.use(express.json());
```
BudnikOleksii, line 5: You can add this middleware to all routes just once
```javascript
app.use(express.json());
```

BudnikOleksii (approved):
👍

PR #576, opened 2024-05-11
Develop

olduma:


src/services/expense.service.js
```diff
@@ -0,0 +1,70 @@
...
+  );
+};
+
+const create = (
+  userId,
+  spentAt = new Date().toISOString(),
+  title,
+  amount = 1,
+  category,
+  note,
```
StasSokolov1, line 24:
```suggestion
  userId,
  title,
  category,
  note,
  amount = 1,
  spentAt = new Date().toISOString(),
```

let's move non required params to the end of list, will be more readable

src/controllers/user.controller.js
```diff
@@ -0,0 +1,92 @@
...
+};
+
+const remove = async (req, res) => {
+  const { id } = req.params;
+
+  try {
+    const userExists = await userService.getUserById(id);
+
+    if (!userExists) {
+      return res.status(404).send({ error: 'User not found' });
```
StasSokolov1, line 52: will be better create const with statuses

StasSokolov1 (approved):
Looks good to me, with only a few minor comments, I'm approving it

PR #578, opened 2024-05-14
solution

ChrisHryts:


src/services/expense.services.js
```diff
@@ -0,0 +1,75 @@
...
+    filteredExpenses = filteredExpenses.filter((expense) => {
+      return new Date(expense.spentAt) >= new Date(from);
+    });
+  }
+
+  if (to) {
+    filteredExpenses = filteredExpenses.filter((expense) => {
+      return new Date(expense.spentAt) <= new Date(to);
+    });
+  }
```
StasSokolov1, line 35: you use filter at least three times here. would be good to combine it to iterate only once

StasSokolov1 (changes requested):
Looks good to me, let's improve your code a bit more

BudnikOleksii (approved):
👍

PR #580, opened 2024-05-14
Solution

Nirvacsh:


src/controllers/expensesController.js
```diff
@@ -0,0 +1,106 @@
...
+  if (
+    !user ||
+    !spentAt ||
+    typeof spentAt !== 'string' ||
+    !title ||
+    typeof title !== 'string' ||
+    !amount ||
+    typeof amount !== 'number' ||
+    !category ||
+    typeof category !== 'string'
```
Moroz-Dmytro, line 43: Create separate function for validation

src/controllers/expensesController.js
```diff
@@ -0,0 +1,106 @@
...
+    return res.sendStatus(statusCode.BAD_REQUEST);
+  }
+
+  const newExpense = expensesService.create(
+    userId,
+    spentAt,
+    title,
+    amount,
+    category,
+    note,
```
Moroz-Dmytro, line 54: Why not to pass all object instead of separate arguments

src/services/expensesService.js
```diff
@@ -0,0 +1,79 @@
...
+    filteredExpenses = filteredExpenses.filter(
+      (item) => item.category === categories,
+    );
+  }
+
+  if (from || to) {
+    filteredExpenses = filteredExpenses.filter(
+      (item) => item.spentAt >= from && item.spentAt <= to,
+    );
+  }
```
Moroz-Dmytro, line 26: DRY. Use `filteredExpenses = filteredExpenses.filter` once and move all `if`s inside

src/services/usersService.js
```diff
@@ -0,0 +1,41 @@
...
+  users = [];
+};
+
+const getAll = () => users;
+
+const getById = (id) => users.find((user) => user.id === Number(id)) || null;
+
+const create = (name) => {
+  const user = {
+    id: users.length,
```
Moroz-Dmytro, line 13: Create function for Id generation, it should be unique in array, because now this value can not be unique

src/utils/statusCodes.js
```diff
@@ -0,0 +1,13 @@
+const SUCCESS = 200;
+const CREATED = 201;
+const NO_CONTENT = 204;
+const BAD_REQUEST = 400;
+const NOT_FOUND = 404;
```
Moroz-Dmytro, line 5: Make it as object `STATUS_CODE` with different status codes

Moroz-Dmytro (changes requested):

PR #581, opened 2024-05-14
task solution

Oleksandr2108:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,87 @@
...
+    !user ||
+    !spentAt ||
+    typeof spentAt !== 'string' ||
+    !title ||
+    typeof title !== 'string' ||
+    !amount ||
+    typeof amount !== 'number' ||
+    !category ||
+    typeof category !== 'string'
+  ) {
```
BudnikOleksii, line 39: Create a separate function for validation

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,87 @@
...
+    return res.sendStatus(400);
```
BudnikOleksii, line 40: Create a constant with status codes

src/services/expenses.services.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  if (categories) {
+    filteredExpenses = filteredExpenses.filter(
+      (item) => item.category === categories,
+    );
+  }
+
+  if (from) {
+    filteredExpenses = filteredExpenses.filter((item) => {
```
BudnikOleksii, line 27: Create a function that filters expenses and try to do it in one filter instead

src/utils/getMaxId.js
```diff
@@ -0,0 +1,13 @@
+function getCreateMaxId(array) {
+  if (array.length === 0) {
+    return 1;
+  } else {
+    const maxId = Math.max(...array.map((obj) => obj.id));
+
+    return maxId + 1;
```
BudnikOleksii, line 7: You've added the `uuid` library, you can use it to generate unique id

BudnikOleksii (changes requested):
Good work 👍 
Left a few comments, let's improve your solution

src/utils/getMaxId.js
Oleksandr2108: At first I made a uuid library, but the tests skip only digits.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,101 @@
+const expensesService = require('../services/expenses.services');
+const usersService = require('../services/users.services');
+
+const STATUS_CODES = {
+  OK: 200,
```
BudnikOleksii, line 5: Move to constants file/folder, keys should be in lowercase

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,101 @@
...
+  CREATED: 201,
+  BAD_REQUEST: 400,
+  NOT_FOUND: 404,
+  NO_CONTENT: 204,
+};
+
+const validateExpense = ({ userId, spentAt, title, amount, category }) => {
```
BudnikOleksii, line 12: Move to utils/validation file/folder

src/controllers/users.controller.js
```diff
@@ -0,0 +1,71 @@
+const usersService = require('../services/users.services');
+
+const getAllUsers = (req, res) => {
+  const users = usersService.getAll();
+
+  res.status(200).json(users);
```
BudnikOleksii, line 6: Use your constant with status codes, fix in all places

src/utils/getMaxId.js
BudnikOleksii: > At first I made a uuid library, but the tests skip only digits.

Ok, so remove the `uuid` package if you don't use it

BudnikOleksii (changes requested):
Good progress, let's improve your solution slightly

StasSokolov1 (approved):
Looks good to me🔥

PR #583, opened 2024-05-14
Add task solution

vladarskyi:


src/controllers/expenseController.js
```diff
@@ -0,0 +1,168 @@
+const expenseService = require('../services/expenseService');
+const { getUserById } = require('../services/userService');
+const { STATUS_CODE } = require('../utils/statusCodes');
+
+const getAll = (req, res) => {
+  const { userId, categories, from, to } = req.query;
```
dffUqp, line 6: You don't really need destructuring here

src/controllers/expenseController.js
```diff
@@ -0,0 +1,168 @@
...
+  const id = Number(userId);
+
+  const AllExpenses = expenseService.getAllExpenses(id, categories, from, to);
```
dffUqp, line 9: Why did you capitalize the name of this variable?

src/controllers/expenseController.js
```diff
@@ -0,0 +1,168 @@
...
+    res.sendStatus(STATUS_CODE.NOT_FOUND);
+
+    return;
+  }
+
+  if (!note || typeof note !== 'string') {
+    res.sendStatus(STATUS_CODE.NOT_FOUND);
+
+    return;
+  }
```
dffUqp, line 58: To handle multiple conditions we basically use switch/case.

Also, you can try to create some variable like `isValidBody` and write just one if statement

src/services/expenseService.js
```diff
@@ -0,0 +1,71 @@
+/* eslint-disable function-paren-newline */
```
dffUqp, line 1: Can we fix this eslint issue?

src/services/expenseService.js
```diff
@@ -0,0 +1,71 @@
...
+const updateExpense = (fieldsToUpdate) => {
+  const expense = expenses.find((usr) => usr.id === fieldsToUpdate.id) || null;
+
+  Object.assign(expense, fieldsToUpdate);
+
+  return expense;
+};
+
+const removeExpense = (id) => {
+  expenses = expenses.filter((usr) => usr.id !== Number(id));
```
dffUqp, line 61:
```suggestion
  expenses = expenses.filter((expense) => expense.id !== Number(id));
```

src/services/expenseService.js
```diff
@@ -0,0 +1,71 @@
...
+
+  if (id) {
+    preparedExpenses = preparedExpenses.filter(({ userId }) => userId === id);
+  }
+
+  if (categories) {
+    preparedExpenses = preparedExpenses.filter(({ category }) =>
+      categories.includes(category),
+    );
+  }
```
dffUqp, line 21: You use filter at least three times here. is it possible to combine it to iterate only once?

src/services/userService.js
```diff
@@ -0,0 +1,47 @@
+const { createUniqueID } = require('../utils/createUniqueID');
+
+let users = [];
+
+const initUsers = () => {
+  users = [];
+};
+
+const getAllUsers = () => {
```
dffUqp, line 9:
```suggestion
const getAll= () => {
```

src/services/userService.js
```diff
@@ -0,0 +1,47 @@
...
+  return users;
+};
+
+const getUserById = (id) => {
```
dffUqp, line 13: You use like userService.[your method] so we already know that it related to users

```suggestion
const getById = (id) => {
```

src/utils/createUniqueID.js
```diff
@@ -0,0 +1,5 @@
+const createUniqueID = () => {
+  return Math.round(Math.random() * Date.now());
+};
```
dffUqp, line 3:
```suggestion
const createUniqueID = () => Math.round(Math.random() * Date.now())

```

dffUqp (changes requested):
Almost done🙂

src/controllers/expenseController.js
vladarskyi: I didn't catch how can I use switch/case here, but made a variable.

src/services/expenseService.js
vladarskyi: Seems like an eslint rules conflict...
Screenshot 2024-05-16 at 13 14 23
>>>

StasSokolov1 (approved):
Looks like a masterpiece, I like it🔥

PR #584, opened 2024-05-14
Added solution

ViktoriiaRepo:


src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
```
dffUqp, line 1: Do we really need it?

src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,68 @@
...
+
+const expensesService = require('../services/expenses.services');
+
+const getExpenses = (req, res) => {
```
dffUqp, line 5: You use it as `expenseConrollers.[your_method]` so we already know that it related to expense

```suggestion
const get = (req, res) => {
```

src/routes/expenses.routes.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const express = require('express');
+
+const expenseConrollers = require('../controllers/expenses.controllers');
+
+const expensesRouter = express.Router();
+
+expensesRouter.get('/', expenseConrollers.getExpenses);
```
dffUqp, line 9:
```suggestion
expensesRouter.get('/', expensesConroller.getExpenses);
```

src/controllers/expenses.controllers.js
```diff
@@ -0,0 +1,68 @@
...
+  const { userId, categories, from, to } = req.query;
```
dffUqp, line 6: You don't really need destructuring here

src/routes/user.routes.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const express = require('express');
+
+const userControllers = require('../controllers/user.controllers');
+
+const userRouter = express.Router();
+
+userRouter.get('/', userControllers.getAllUsers);
```
dffUqp, line 9:
```suggestion
userRouter.get('/', userController.getAllUsers);
```

src/services/expenses.services.js
```diff
@@ -0,0 +1,105 @@
...
+    filteredExpenses = filteredExpenses.filter(
+      (expense) => new Date(expense.spentAt) >= new Date(from),
+    );
+  }
+
+  if (to) {
+    filteredExpenses = filteredExpenses.filter(
+      (expense) => new Date(expense.spentAt) <= new Date(to),
+    );
+  }
```
dffUqp, line 36: You use filter at least three times here. is it possible to combine it to iterate only once?

dffUqp (changes requested):
Almost done🙂

src/controllers/user.controllers.js
```diff
@@ -0,0 +1,86 @@
+const userService = require('../services/user.services');
+
+const getAll = (req, res) => {
+  if (!userService.getAll()) {
```
StasSokolov1, line 4: you call it twice, let's create variable like `allUsers`

PR #586, opened 2024-05-14
dev: add solution

tetlisna:


src/routes/expenses.routes.js
```diff
@@ -0,0 +1,89 @@
+const express = require('express');
+const router = express.Router();
+
+const createRouter = ({ userService, expensesService }) => {
```
dffUqp, line 4: 1. You can directly import your services into this file 
2. It is enough you need to export only the router variable from this file

src/routes/expenses.routes.js
```diff
@@ -0,0 +1,89 @@
...
+      returnData = returnData.filter((e) => e.category === categories);
+    }
+
+    if (from && to) {
+      returnData = returnData.filter(
+        (e) => e.spentAt >= from && e.spentAt < to,
+      );
+    }
+
+    res.send(returnData);
```
dffUqp, line 26: It is better to store such logic in separate controller file

src/routes/expenses.routes.js
```diff
@@ -0,0 +1,89 @@
...
+    if (!expense) {
+      return res.sendStatus(404);
+    }
+    res.send(expense);
+  });
+
+  router.post('/', (req, res) => {
+    const { title, spentAt, userId, amount, category, note } = req.body;
+
+    if (!title || !amount || !category || !spentAt || !userId) {
```
dffUqp, line 42: Create a separate variable for this

src/routes/expenses.routes.js
```diff
@@ -0,0 +1,89 @@
...
+
+  router.delete('/:id', (req, res) => {
+    const { id } = req.params;
+    const deleted = expensesService.deleteOne(id);
+
+    if (!deleted) {
+      return res.sendStatus(404);
+    }
+
+    res.sendStatus(204);
```
dffUqp, line 83: Create constants for all status codes

dffUqp (changes requested):
Almost done🙂

PR #588, opened 2024-05-15
add solution

Math8585:


src/controller/expense.controller.js
```diff
@@ -0,0 +1,74 @@
+const expenseServise = require('../services/expenseService');
+const usersService = require('../services/userService');
+
+const expenseController = {};
+
+expenseController.get = (req, res) => {
```
Moroz-Dmytro, line 6:
```suggestion
const getAll = (req, res) => {
```

The same with all other functions

src/controller/expense.controller.js
```diff
@@ -0,0 +1,74 @@
...
+
+    return;
+  }
+
+  expenseServise.remove(id);
+
+  res.sendStatus(204);
+};
+
+module.exports = expenseController;
```
Moroz-Dmytro, line 74:
```suggestion
module.exports = {
  getAll,
  .......
};
```

src/controller/expense.controller.js
```diff
@@ -0,0 +1,74 @@
+const expenseServise = require('../services/expenseService');
+const usersService = require('../services/userService');
+
+const expenseController = {};
+
+expenseController.get = (req, res) => {
+  const { userId, categories, from, to } = req.query;
```
Moroz-Dmytro, line 7: You don't really need destructuring here

src/controller/expense.controller.js
```diff
@@ -0,0 +1,74 @@
...
+  const filteredExpenses = expenseServise.getAll(userId, categories, from, to);
+
+  res.json(filteredExpenses);
+};
+
+expenseController.getOne = (req, res) => {
```
Moroz-Dmytro, line 13:
```suggestion
const getById = (req, res) => {
```

src/controller/expense.controller.js
```diff
@@ -0,0 +1,74 @@
...
+  const { id } = req.params;
+
+  const expense = expenseServise.getById(id);
+
+  if (!expense) {
+    res.sendStatus(404);
```
Moroz-Dmytro, line 19: Create variable for all status codes

src/controller/user.controller.js
```diff
@@ -0,0 +1,70 @@
+const usersService = require('../services/userService');
+
+const userController = {};
+
+userController.get = (req, res) => {
```
Moroz-Dmytro, line 5: The same comments as for `expensesController`

src/services/expenseService.js
```diff
@@ -0,0 +1,81 @@
...
+
+expenseServise.reset = () => {
+  expenses.length = 0;
+};
+
+expenseServise.getAll = (userId, categories, from, to) => {
+  let filteredExpenses = expenses;
+
+  if (userId) {
+    filteredExpenses = filteredExpenses.filter(
```
Moroz-Dmytro, line 15: DRY, use `filteredExpenses = filteredExpenses.filter(` only once, and move all `if`s inside it

src/services/expenseService.js
```diff
@@ -0,0 +1,81 @@
...
+
+  return filteredExpenses;
+};
+
+expenseServise.getById = (id) => {
+  return expenses.find((outlay) => outlay.id === Number(id)) || null;
+};
+
+expenseServise.create = (body) => {
+  const { userId, title, amount, category, note, spentAt } = body;
```
Moroz-Dmytro, line 50: You don't need destructuring here

src/services/expenseService.js
```diff
@@ -0,0 +1,81 @@
...
+
+  const expense = {
+    id: Math.floor(Math.random() * 10000),
```
Moroz-Dmytro, line 53: Create function for id generation. This logic is not really good, because `id` can be repeated

src/services/userService.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+userServise.getById = (id) => {
+  return users.find((person) => person.id === Number(id)) || null;
+};
+
+userServise.create = (name) => {
+  const user = {
+    name,
+    id: Math.floor(Math.random() * 10000),
```
Moroz-Dmytro, line 20: Same comment as above

Moroz-Dmytro (changes requested):
Good progress 👍 
But check some comments below
Also, try to check other places for similar mistakes, to fix them in all places

BudnikOleksii (approved):
👍

PR #589, opened 2024-05-15
add solution

illya-onyshchuk:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,90 @@
...
+const userService = require('../services/user.service');
+
+const getAll = (req, res) => {
+  const { userId, categories, from, to } = req.query;
+
+  res.statusCode = 200;
+  res.send(getExpenses(userId, categories, from, to));
+};
+
+const getExpense = (req, res) => {
```
dffUqp, line 17: You use like expenseController.[your method] so we already know that it related to expense
```suggestion
const getById = (req, res) => {
```

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,90 @@
+const {
+  getExpenses,
+  getExpenseById,
+  create,
+  remove,
+  update,
+} = require('../services/expenses.service');
```
dffUqp, line 7: I think you don't need destructuring here

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,90 @@
...
+const updateExpense = (req, res) => {
+  const { id } = req.params;
+  const body = req.body;
+  const expense = getExpenseById(+id);
+
+  if (!expense) {
+    return res.sendStatus(404);
+  }
+
+  res.statusCode = 200;
```
dffUqp, line 80: It will be nice to create constants for all status codes

dffUqp (approved):
LGTM 🐸

PR #590, opened 2024-05-15
Solution to accounting-app

winniepuhh:


src/controllers/expenses.contr.js
```diff
@@ -0,0 +1,68 @@
+const expensesServ = require('../services/expenses.service');
```
dffUqp, line 1: Why don't you use a full name expensesService. 

I think now it looks a bit weird

src/controllers/expenses.contr.js
```diff
@@ -0,0 +1,68 @@
...
+const usersServ = require('../services/users.service');
```
dffUqp, line 2: Same about naming

src/controllers/expenses.contr.js
```diff
@@ -0,0 +1,68 @@
...
+const statusCode = require('../utils/statusCode');
+
+const getAllExpenses = (req, res) => {
```
dffUqp, line 5: You use like `expensesContr.[your method]` so we already know that it related to expenses
```suggestion
const getAll = (req, res) => {
```

src/services/expenses.service.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable function-paren-newline */
```
dffUqp, line 1: Can we fix this eslint issue? 

(if we can't, let's disable it for the whole app using the config)

dffUqp (approved):
LGTM😶‍🌫️

PR #591, opened 2024-05-15
add accounting app

DenNice-cloud:
## Description
- Work on accounting app with "Express"

### DIFF
--- a/src/controllers/expense.controller.js
+++ b/src/controllers/expense.controller.js
@@ -1,0 +1,77 @@
+const expenseService = require('../services/expense.service');
+const userService = require('../services/user.service');
+
+const getAll = (req, res) => {
+  res.statusCode = 200;
+  res.send(expenseService.getAll(req.query));
+};
+
+const getById = (req, res) => {
+  const { id } = req.params;
+  const expense = expenseService.getById(id);
+
+  if (!expense) {
+    return res.sendStatus(404);
+  }
+
+  res.statusCode = 200;
+  res.send(expense);
+};
+
+const postExpense = (req, res) => {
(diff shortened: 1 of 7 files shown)


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  if (
+    !spentAt ||
+    !title ||
+    typeof title !== 'string' ||
+    !amount ||
+    typeof amount !== 'number' ||
+    !category ||
+    typeof category !== 'string'
```
dffUqp, line 35: it will be better to create a separate variable for this

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,77 @@
...
+
+  res.statusCode = 200;
+  res.send(expense);
+};
+
+const postExpense = (req, res) => {
+  const { userId, spentAt, title, amount, category } = req.body;
+
+  if (!userService.getById(userId)) {
+    return res.sendStatus(400);
```
dffUqp, line 25: Let's create constants for all status codes

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  const expense = expenseService.create(req.body);
+
+  res.statusCode = 201;
+  res.send(expense);
+};
+
+const deleteExpense = (req, res) => {
```
dffUqp, line 46: You use like `expenseController.[your method]` so we already know that it related to expense
```suggestion
const delete = (req, res) => {
```

dffUqp (approved):
LGTM🐏

PR #595, opened 2024-05-17
add solution

Vlobdam:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,99 @@
+const expensesService = require('../services/expenses.service');
+const usersService = require('../services/users.service');
+
+const getAll = async (req, res) => {
+  const { from, to, userId, categories } = req.query;
```
Moroz-Dmytro, line 5: You don't need destructuring here

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,99 @@
...
+  );
+};
+
+const getById = async (req, res) => {
+  const { id } = req.params;
+
+  const expense = expensesService.getById(id);
+
+  if (!expense) {
+    res.sendStatus(404);
```
Moroz-Dmytro, line 23: Create variable for all status codes

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,99 @@
...
+const create = async (req, res) => {
+  const { userId, title, amount, category, note, spentAt } = req.body;
+
+  if (!userId || !title || !amount || !category || !spentAt) {
+    res.sendStatus(400);
+
+    return;
+  }
+
+  if (!usersService.getById(userId)) {
```
Moroz-Dmytro, line 40: Can you move this check inside previous `if`?
You return the same result in both of them

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,99 @@
...
+
+  res.send(
+    expensesService.create({
+      userId,
+      title,
+      amount,
+      category,
+      note,
+      spentAt,
+    }),
```
Moroz-Dmytro, line 56:
```suggestion
    expensesService.create(req.body),
```

src/controllers/users.controller.js
```diff
@@ -0,0 +1,76 @@
...
+
+  res.statusCode = 201;
+  res.send(usersService.create({ name }));
+};
+
+const getById = async (req, res) => {
+  const { id } = req.params;
+  const user = usersService.getById(id);
+
+  if (user === undefined) {
```
Moroz-Dmytro, line 24:
```suggestion
  if (!user) {
```

src/services/expenses.service.js
```diff
@@ -0,0 +1,85 @@
...
+  expenses = [];
+
+  return expenses;
+};
+
+const getAll = ({ from, to, userId, categories }) => {
+  let filteredExpenses = expenses;
+
+  if (from) {
+    filteredExpenses = filteredExpenses.filter((expense) => {
```
Moroz-Dmytro, line 14: DRY. Use `filteredExpenses = filteredExpenses.filter((expense) => {` only onde, and move all `if`s inside it

src/services/expenses.service.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  return filteredExpenses;
+};
+
+const getById = (id) => {
+  return expenses.find((expense) => expense.id === Number(id));
+};
+
+const create = ({ userId, spentAt, title, amount, category, note }) => {
```
Moroz-Dmytro, line 48: You don't need this destructuring

src/services/users.service.js
```diff
@@ -0,0 +1,50 @@
...
+
+  Object.assign(user, { name });
+
+  return user;
+};
+
+const remove = (id) => {
+  const index = getById(id);
+
+  users.splice(index, 1);
```
Moroz-Dmytro, line 40: Better use `filter`

src/utils/newId.js
```diff
@@ -0,0 +1,12 @@
+const newId = (arr) => {
```
Moroz-Dmytro, line 1:
```suggestion
const getNewId = (arr) => {
```

Moroz-Dmytro (changes requested):

BudnikOleksii (approved):
👍

PR #596, opened 2024-05-17
add task solution

dima8528:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,84 @@
+const expenseService = require('./../services/expense.service');
+const userService = require('./../services/user.service');
+
+const get = (req, res) => {
```
Moroz-Dmytro, line 4:
```suggestion
const getAll = (req, res) => {
```

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,84 @@
...
+  const expenses = expenseService.get(req.query);
+
+  res.status(200).send(expenses);
```
Moroz-Dmytro, line 7: Create variable for all status codes

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,84 @@
...
+  const elements = ['spentAt', 'title', 'amount', 'category', 'note'];
+  const values = {};
+
+  elements.forEach((el) => {
+    const value = req.body[el];
+
+    if (value) {
+      values[el] = value;
+    }
+  });
```
Moroz-Dmytro, line 58: Why do you need this?
Why not to just use `req.body`

src/controllers/user.controller.js
```diff
@@ -0,0 +1,74 @@
+const userService = require('./../services/user.service');
+
+const get = (_, res) => {
```
Moroz-Dmytro, line 3:
```suggestion
const getAll = (_, res) => {
```

src/services/expense.service.js
```diff
@@ -0,0 +1,78 @@
...
+
+const createNewId = () => {
+  if (!expenses.length) {
+    return 1;
+  }
+
+  const expensesIdsArr = expenses.map((expense) => expense.id);
+
+  return Math.max(...expensesIdsArr) + 1;
+};
```
Moroz-Dmytro, line 16: Move this function to separate file and reuse in both services

src/services/expense.service.js
```diff
@@ -0,0 +1,78 @@
...
+
+const get = ({ userId, from, to, categories }) => {
```
Moroz-Dmytro, line 18:
```suggestion
const getAll = ({ userId, from, to, categories }) => {
```

Moroz-Dmytro (changes requested):

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,74 @@
+const statusCodes = require('./../utils/statusCodes');
+const expenseService = require('./../services/expense.service');
+const userService = require('./../services/user.service');
+
+const getAll = (req, res) => {
+  const expenses = expenseService.getAll(req.query);
+
+  res.status(statusCodes['OK']).send(expenses);
```
BudnikOleksii, line 8:
```suggestion
  res.status(statusCodes.OK).send(expenses);
```
Use square brackets when you need to access with a dynamic key, in this case, you can access with a dot, fix in all places

src/services/user.service.js
```diff
@@ -0,0 +1,44 @@
+const { getMaxID } = require('../utils/getMaxID');
+// console.log(getMaxID);
```
BudnikOleksii, line 2: Remove comments

src/utils/statusCodes.js
```diff
@@ -0,0 +1,14 @@
+const statusCodes = {
+  OK: 200,
```
BudnikOleksii, line 2:
```suggestion
const STATUS_CODES = {
  ok: 200,
```
Should be vise versa, constant name in UPPER_CASE and keys in camelCase

BudnikOleksii (changes requested):
Good job 👍 
Just a few minor fixes are required

BudnikOleksii (approved):
👍

PR #598, opened 2024-05-20
serve API

Ihor-Prodan:


src/createServer.js
```diff
@@ -1,11 +1,193 @@
...
+
+    res.status(201).json(newUser);
+  });
+
+  app.get('/users', (req, res) => {
+    res.status(200).json(users);
+  });
+
+  app.get('/users/:id', (req, res) => {
+    const user = users.find((u) => u.id === parseInt(req.params.id));
```
DarkMistyRoom, line 35:
```suggestion
    const user = users.find((user) => user.id === parseInt(req.params.id));
```

DarkMistyRoom: don't use one-letter variables

src/createServer.js
```diff
@@ -1,11 +1,193 @@
...
+    const { name } = req.body;
+    const user = users.find((u) => u.id === parseInt(req.params.id));
+
+    if (!user) {
+      return res.status(404).json({ message: 'User not found' });
+    }
+
+    if (!name) {
+      return res.status(400).json({ message: 'Missing required parameters' });
+    }
```
DarkMistyRoom, line 53: it makes no sense to find a user if the name wasn't passed

src/createServer.js
```diff
@@ -1,11 +1,193 @@
...
+      filteredExpenses = filteredExpenses.filter(
+        (e) => new Date(e.spentAt) >= new Date(from),
+      );
+    }
+
+    if (to) {
+      filteredExpenses = filteredExpenses.filter(
+        (e) => new Date(e.spentAt) <= new Date(to),
+      );
+    }
```
DarkMistyRoom, line 128: You don't need so many iterations. Combine all conditions into one filter callback

DarkMistyRoom (changes requested):
Create a structure of your code. Make routes, controllers, and services

src/createServer.js
```diff
@@ -1,11 +1,165 @@
...
+
+    if (!userGet) {
+      return res.status(404).json({ message: 'User not found' });
+    }
+    res.status(200).json(userGet);
+  });
+
+  app.patch('/users/:id', (req, res) => {
+    const { name } = req.body;
+    const userPatch = users.find((user) => user.id === parseInt(req.params.id));
```
DarkMistyRoom, line 45: i meant to move `if (!name)` before finding a user

DarkMistyRoom (changes requested):
This comment is still not fixed. Consider using the same structure as in the course videos
image

lerastarynets (approved):
lgtm 🚀

PR #599, opened 2024-05-24
solution

ArtemZakhar:
Hi)

I have several problems with tests. 
One of them I don't understand what I need to do, the rest - I don't want to change anything because I like my approach more and I have permission from the mentor that I can leave it as it is)

I save all data in JSON - that is why tests are failed.

And the last one test: 
image

I tested several times and all looks great, but the mistake is still there(((

image


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,122 @@
...
+  const newExpense = await updateExpenseById(+id, data);
+
+  if (!newExpense) {
+    return res
+      .status(response[404].statusCode)
+      .json({ success: false, error: response[404].messages.noExpense });
+  }
+
+  return res.status(response[200].statusCode).json(newExpense);
+}
```
DanilWeda, line 114: As I see you update all expenses in case when expenses are not found. 
Check the comment below.

src/models/expenses.model.js
```diff
@@ -0,0 +1,107 @@
...
+  if (!updatedExpense) {
+    return;
+  }
+
+  const fullPath = path.join(__dirname, '..', 'db', 'expenses.json');
+
+  fs.writeFileSync(fullPath, JSON.stringify({ expenses }));
+
+  return updatedExpense;
+}
```
DanilWeda, line 99: In this case, you can try to use find and search currentExpense, then if an expense not found you return falsy value ( or nothing) without updating. If you find an expense map element update it and rewrite values.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,122 @@
...
+  const data = req.body;
+  const { id } = req.params;
+
+  if (!Object.values(data).length) {
+    return res
+      .status(response[400].statusCode)
+      .json({ success: false, error: response[400].messages.noData });
+  }
+
+  const newExpense = await updateExpenseById(+id, data);
```
DanilWeda, line 105:
```suggestion
  const newExpense = await updateExpenseById(Number(id), data);
```
as alternative

DanilWeda (changes requested):
Good job. Check some comments that i left. 
Also, you often use let in cases where it is not critically needed. 
Try to use const as default, and then if you don't see another way - use const. 
Thanks)

src/models/expenses.model.js
ArtemZakhar: I just wanted to make it in one itteration)
so I will leave it as it is)

src/controllers/expenses.controller.js
ArtemZakhar: Please clarify what do you mean?)

ArtemZakhar:
My prettier change all let to const in case where it needs automatically) 
I'm pretty sure it has to be all correct about this)

src/controllers/expenses.controller.js
DanilWeda: I about the comment which I left for `updateExpenseById`  method)

DanilWeda (approved):
Okey, I approved this request with the comments which I left.

src/models/expenses.model.js
ArtemZakhar: if there is no expense I make this check) 
It would be undefined if no expense is found)

` if (!updatedExpense) {
    return;
  }`

PR #606, opened 2024-06-14
Node accouting app v2

itsXenthis:


src/createServer.js
```diff
@@ -1,11 +1,22 @@
 'use strict';
 
-// const express = require('express');
+const cors = require('cors');
+const express = require('express');
+const { userRouter } = require('./routes/user.route');
+const userService = require('./services/user.service');
+const expenseService = require('./services/expenses.service');
+const { expenseRouter } = require('./routes/expenses.route');
+const app = express();
```
nataliaklonowska, line 9: why is `const app = express();` outside of createServer function?

nataliaklonowska (approved):
Great work 🔥

PR #607, opened 2024-06-14
Develop2

marinatea:


choeqq (changes requested):
@marinatea please make sure the tests are working

marinatea:
the problem was in "all works good with part of users", but tests doesn't pass. so on the last q&a was agry to I put on gitHub, to mentors check where it can be a mistake :)

marinatea:
> @marinatea please make sure the tests are working

the problem was in "all works good with part of users", but tests doesn't pass. so on the last q&a was agry to I put on gitHub, to mentors check where it can be a mistake :)

src/controllers/expenses.controler.js
choeqq, line 1: why is this file commented out?

src/routers/expenses.router.js
choeqq, line 1: why is this file commented out?

src/services/expenses.service.js
choeqq, line 1: why is this file commented out?

choeqq (changes requested):

src/controllers/expenses.controler.js
marinatea: I did not make expenses part yet. just prepared something and commented because I do not use it yet anywhere.

all going around part of users now.
part of users does not pass 3 tests and I can't fid a mistake. because practically all works.

marinatea: like for exumple: test shows: "User › deleteUser › should return 404 if user does not exist. Expected 404 'Not Found', got 204 'No Content' ". But when I try to delete user which is not exist ,it return me 404, not 204. So I do not understand where can be a mistake. The same with the rest 2 tests.

PR #617, opened 2024-06-25
Solution

VitaliiNez:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,101 @@
...
+  getExpenseById,
+  getExpenseByIdIndex,
+  removeExpenseById,
+  updateExpenseById,
+} = require('../services/expenses.service');
+
+const { getUserById } = require('../services/users.service');
+
+const getFilteredExpense = (req, res) => {
+  const { userId, from, to, categories } = req.query;
```
anastasiiavorobiova, line 13: Consider creating a folder for middleware and validate the user request before passing it to the controller. Also, you can throw the 400 error here

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,101 @@
...
+
+  const expense = getFilteredExpenses({
+    userId,
+    from,
+    to,
+    categories,
+  });
+
+  return res.status(200).send(expense);
```
anastasiiavorobiova, line 22:
```suggestion
try {
  const expense = getFilteredExpenses({
    userId,
    from,
    to,
    categories,
  });

  return res.status(200).send(expense);
}
catch (err) {
  res.status(500).send('some general error message');
}
```

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const setExpense = (req, res) => {
+  const { userId, spentAt, title, amount, category, note } = req.body;
```
anastasiiavorobiova, line 26: The same issue here

anastasiiavorobiova (changes requested):
Great work! Almost done

PR #620, opened 2024-07-04
task solution

kinqbert:


src/controllers/expense.controller.js
```diff
@@ -0,0 +1,93 @@
...
+  };
+
+  const user = User.getUser(userId);
+
+  let propsAreValid = true;
+
+  for (const key in properties) {
+    // can't add just !properties[key], because prop could be 0
+    if (
+      properties[key] === undefined ||
```
mrkovaliv, line 27: Move this condiotion to variable and use here

src/controllers/expense.controller.js
```diff
@@ -0,0 +1,93 @@
...
+      properties[key] === null ||
+      properties[key] === ''
+    ) {
+      propsAreValid = false;
+      break;
+    }
+  }
+
+  if (!propsAreValid || !user) {
```
mrkovaliv, line 36: the same

mrkovaliv (approved):
Good job 👍 👍 👍

PR #621, opened 2024-07-04
Solution

MatheusStefan:


.editorconfig
pedro-ruas, line 1: There's no need in committing this file, it's for your own machine usage

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+  const mockExpenses = [];
+
+  app.use(cors());
+  app.use(express.urlencoded({ extended: true }));
+  app.use(express.json());
+  app.use('/', router);
+
+  app.get('/users', (req, res) => {
+    if (mockUsers.length === 0) {
+      return res.send([]);
```
pedro-ruas, line 20: Is this _if_ statement needed?
Removing it will also send(mockUsers) which is an empty array by default

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+    res.status(200).send(mockUsers);
+  });
+
+  app.post('/users', (req, res) => {
+    const { name } = req.body;
+
+    if (!name) {
+      res.status(400).send('Name is required!');
+
+      return;
```
pedro-ruas, line 31: You can return the res itself, just move this _return_ to the beginning of line 29

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+    }
+
+    const maxId = mockUsers.reduce(
+      (max, user) => (user.id > max ? user.id : max),
+      0,
+    );
+    const newId = maxId + 1;
```
pedro-ruas, line 38: This line can go inside the reduce itself
```
const newId = mockUsers.reduce(
    (max, user) => (user.id > max ? user.id + 1 : max),
    0,
):
```

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+
+  app.get('/users/:id', (req, res) => {
+    const { id } = req.params;
+
+    const newUser = mockUsers.find((user) => user.id === Number(id));
+
+    if (!newUser) {
+      res.sendStatus(404);
+
+      return;
```
pedro-ruas, line 58: Again, you can return the res

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+    }
+
+    const userIndex = mockUsers.findIndex((user) => user.id === Number(id));
+
+    if (userIndex === -1) {
+      res.sendStatus(404);
+
+      return;
```
pedro-ruas, line 77: And again

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+const router = express.Router();
+const cors = require('cors');
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+  const mockUsers = [];
+  const mockExpenses = [];
```
pedro-ruas, line 11: mockUsers and mockExpenses could both be any type o hashmap instead of arrays.
meaning, both Object and Map classes could handle better requests for finding, deleting and so on

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+
+    const newUser = mockUsers.find((user) => user.id === Number(id));
+
+    if (!newUser) {
+      res.sendStatus(404);
+
+      return;
+    }
+
+    if (typeof name !== 'string') {
```
pedro-ruas, line 96: You could join both _if_ statements in one using a _||_

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+      filteredExpenses = filteredExpenses.filter((expense) => {
+        const expenseDate = new Date(expense.spentAt);
+        const startDateDate = new Date(from);
+        const toDate = new Date(to);
+
+        return expenseDate >= startDateDate && expenseDate <= toDate;
+      });
+    }
+
+    if (filteredExpenses.length === 0) {
```
pedro-ruas, line 140: Do you need this _if_ ?

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+        .map((category) => category.trim());
+
+      if (categoryList.length > 0) {
+        filteredExpenses = filteredExpenses.filter((expense) =>
+          categoryList.includes(expense.category),
+        );
+      }
+    }
+
+    if (from && to) {
```
pedro-ruas, line 130: All these _if_ approaches work, but could become something hard to maintain and scale in case of more and more properties/queries needed.
This will keep all scalability in a single place and remove multiple _if_ checks. (because your code checks all ifs, even when already passing the first one).

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+
+    if (!user) {
+      return res.status(400).send('User not found!');
+    }
+
+    const maxId = mockExpenses.reduce(
+      (max, expense) => (expense.id > max ? expense.id : max),
+      0,
+    );
+    const newId = maxId + 1;
```
pedro-ruas, line 164: Same as before, you can move this to inside the reduce

src/createServer.js
```diff
@@ -1,11 +1,233 @@
...
+    const { id } = req.params;
+    const { title, amount, category, note } = req.body;
+
+    const expense = mockExpenses.find((item) => item.id === Number(id));
+
+    if (!expense) {
+      return res.sendStatus(404);
+    }
+
+    expense.title = title !== undefined ? title : expense.title;
```
pedro-ruas, line 222: `expense.title = title ?? expense.title`

pedro-ruas (changes requested):

src/controllers/expensesController.js
```diff
@@ -0,0 +1,40 @@
+const Expense = require('./Expense');
+const { mockExpenses } = require('./data');
+
+const getExpenses = () => [...mockExpenses.values()];
+
+const getExpenseById = (id) => mockExpenses.get(id);
+
+const addExpense = (userId, spentAt, title, amount, category, note) => {
+  const maxId = Math.max(...mockExpenses.keys(), 0);
+  const newId = maxId + 1;
```
pedro-ruas, line 10: You can merge both lines into one:
`const newId = Math.max(...mockExpenses.keys(), 0) + 1;`

src/controllers/expensesController.js
pedro-ruas, line 1: This file is _almost_ flawless, but you named it wrong.
A _Controller_ handles HTTP interactions, it receives a request, validates the data, and if all approved, sends the data to a _Service_.
The _Service_ then accesses repositories and uses _Models_ and _Views_ to return the needed data accordingly to the input passed by the _Controller_.

What I mean is, this code works just fine, but it isn't a _Controller_, it looks more like a _Service_.

src/controllers/usersController.js
pedro-ruas, line 1: Check advice from the previous file, they are the same

src/models/Expense.js
```diff
@@ -0,0 +1,13 @@
...
+    this.id = id;
+    this.userId = userId;
+    this.spentAt = spentAt;
+    this.title = title;
+    this.amount = amount;
+    this.category = category;
+    this.note = note;
+  }
+}
+
```
pedro-ruas, line 12: There could be implementations for _getters_ and _setters_ too

src/models/User.js
```diff
@@ -0,0 +1,8 @@
+class User {
+  constructor(id, name) {
+    this.id = id;
+    this.name = name;
+  }
+}
```
pedro-ruas, line 6: Same here, _getters_ and _setters_ would be great to see

src/routes/expensesRouter.js
```diff
@@ -0,0 +1,127 @@
+/* eslint-disable function-paren-newline */
+const { Router } = require('express');
+const { mockUsers } = require('./usersRouter.js');
+const expensesRouter = Router();
+const mockExpenses = new Map();
+
+const filterFunctions = {
```
pedro-ruas, line 7: This filter could be inside your _Service_ file (the file you called _Controller_), would be a better place to store it

src/routes/expensesRouter.js
```diff
@@ -0,0 +1,127 @@
...
+
+  if (userId) {
+    filteredExpenses = filterFunctions.userId(filteredExpenses, userId);
+  }
+
+  if (categories) {
+    filteredExpenses = filterFunctions.categories(filteredExpenses, categories);
+  }
+
+  if (from && to) {
```
pedro-ruas, line 42: Again, this filtering process could be inside your _Service_ file.
Aaaand, there is a better approach for applying those filters. If they are already stored in a Hashmap (_filterFunctions_) you could use a single filter to apply then all, no _if_ statements needed.

src/routes/expensesRouter.js
```diff
@@ -0,0 +1,127 @@
...
+    filteredExpenses = filterFunctions.dateRange(filteredExpenses, from, to);
+  }
+
+  res.status(200).send(filteredExpenses);
```
pedro-ruas, line 46: This type of response looks like a _Controller_

src/routes/expensesRouter.js
pedro-ruas, line 1: Here you have your actual _Controllers_ baked onto your _Routes_.
For every _Route_ declaration here, you are also passing a function to handle it, right? I. e.:
```
expensesRouter.delete('/:id', (req, res) => {
  const { id } = req.params;

  if (!id || isNaN(Number(id))) {
    return res.sendStatus(400);
  }

  if (!mockExpenses.has(Number(id))) {
    return res.sendStatus(404);
  }

  mockExpenses.delete(Number(id));
  res.sendStatus(204);
});
```
Everything after the first argument is a controller that could be in a controller file: "('/:id', <THIS_IS_YOUR_CONTROLLER_FUNCTION)`,

So you could have a Class called ExpenseController with all methods, and you would call here:
```
const expenseController = new ExpenseController();

expensesRouter.delete('/:id', expenseController.delete);
```

src/routes/usersRouter.js
pedro-ruas, line 1: Same advice from previous file

tests/expense.test.js
```diff
@@ -2,12 +2,16 @@
...
+const { resetMockUsers } = require('../src/routes/usersRouter');
+const { resetMockExpenses } = require('../src/routes/expensesRouter');
 
 describe('Expense', () => {
   let server;
   let api;
 
   beforeEach(() => {
+    resetMockExpenses();
+    resetMockUsers();
```
pedro-ruas, line 14: You could actually prevent this if your Repository were a Class.
But that is another topic xD, you don't have to change it here

pedro-ruas (changes requested):

tests/user.test.js
```diff
@@ -2,12 +2,14 @@
 
 const supertest = require('supertest');
 const { createServer } = require('../src/createServer');
+const { resetMockUsers } = require('../src/services/userServices.js');
 
-describe('User', () => {
+describe.skip('User', () => {
```
EdPirro, line 7: You're skipping User tests

EdPirro (changes requested):
Great job, @MatheusStefan! 🚀
The most important issues were fixed and your overall code structure looks way better, but you skipped an entire test suite! Please submit again against all tests.

jv-aquino (approved):
Gj Matheus! There are some OOP principles (like getters and setters that could've been also used but for now that's really good!

PR #622, opened 2024-07-05
add task solution

PrytulaAlexandr:
add task solution

### DIFF
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -1,13 +1,185 @@
 'use strict';
 
-// const express = require('express');
+const express = require('express');
 
 function createServer() {
-  // Use express to create a server
-  // Add a routes to the server
-  // Return the server (express app)
+  const app = express();
+
+  app.use(express.json());
+
+  const users = [];
+  const expenses = [];
+  let nextUserId = 1;
+  let nextExpenseId = 1;
+
+  app.get('/users', (req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+    res.send(users);
+  });
+
+  app.get('/users/:id', (req, res) => {
+    const { id } = req.params;
+    const user = users.find((people) => people.id === parseInt(id));
+
+    if (!user) {
+      res.sendStatus(404);
+
+      return;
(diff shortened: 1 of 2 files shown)


src/services/user.services.js
```diff
@@ -0,0 +1,46 @@
+let users = [];
```
DanilWeda, line 1: You can use const, then make operation with array

src/services/user.services.js
```diff
@@ -0,0 +1,46 @@
...
+
+const getUserByIdService = (id) => {
+  const user = users.find((people) => people.id === +id) || null;
+
+  return user;
+};
+
+const createUserService = (newName) => {
+  const newUser = {
+    id: nextUserId++,
```
DanilWeda, line 16: You can create generateId func for it, and use it in all places in your app

src/services/expense.services.js
```diff
@@ -0,0 +1,98 @@
+let expenses = [];
+let nextExpenseId = 1;
```
DanilWeda, line 2: The same story

src/routes/users.router.js
```diff
@@ -0,0 +1,15 @@
+const express = require('express');
+const userController = require('../constrollers/users.controller.js');
+const usersRouter = express.Router();
+
+usersRouter.get('/', userController.getAllUsers);
```
DanilWeda, line 5: You can create constants for all routes

src/constrollers/users.controller.js
```diff
@@ -0,0 +1,73 @@
...
+
+  res.json(updatedUser);
+};
+
+const deleteUser = (req, res) => {
+  const { id } = req.params;
+  const user = services.getUserByIdService(id);
+
+  if (!user) {
+    res.status(404).json({ error: 'User not found' });
```
DanilWeda, line 59: Move all status and messages in to constants ( for all files)

DanilWeda (changes requested):
Good job. 
Check some small comments. 
Many thanks)

PR #625, opened 2024-07-05
Develop

eduvinagre:


src/controllers/expenseController.js
```diff
@@ -0,0 +1,74 @@
...
+}
+
+function getAllExpenses(req, res) {
+  const expenses = expenseModel.getAllExpenses(req.query);
+
+  res.json(expenses);
+}
+
+function getExpenseById(req, res) {
+  const expense = expenseModel.getExpenseById(parseInt(req.params.id));
```
pedro-ruas, line 39: This function is unnecessary, as it does the same as the previous one, _getAllExpenses_, when passing the right query.
Meaning, the following code replaces this function:
```
const [expense] = expenseModel.getAllExpenses({ id: parseInt(req.params.id) });
```

src/controllers/userController.js
```diff
@@ -0,0 +1,62 @@
...
+}
+
+function getAllUsers(req, res) {
+  const users = userModel.getAllUsers();
+
+  res.json(users);
+}
+
+function getUserById(req, res) {
+  const user = userModel.getUserById(parseInt(req.params.id));
```
pedro-ruas, line 24: You could use a similar solution to what you did for Expenses, where _getAllUsers_ could accept a query and solve all cases at once.
Then you would be able to search users not only by Id, but by any wanted prop

src/models/expenseModel.js
```diff
@@ -0,0 +1,82 @@
+/* eslint-disable no-console */
+/* eslint-disable function-paren-newline */
+/* eslint-disable no-shadow */
+'use strict';
+
+const expenses = [];
+
+const expenseModel = {
```
pedro-ruas, line 8: expenseModel  could be a Class, but that's ok

src/models/expenseModel.js
```diff
@@ -0,0 +1,82 @@
+/* eslint-disable no-console */
+/* eslint-disable function-paren-newline */
+/* eslint-disable no-shadow */
+'use strict';
+
+const expenses = [];
```
pedro-ruas, line 6: IT would be better for performing CRUD operations if _expenses_ were some sort of Hashmap, like an Object or Map instance

src/models/expenseModel.js
```diff
@@ -0,0 +1,82 @@
...
+    const filteringTests = {
+      userId: (id, expense) => Number(id) === expense.userId,
+      id: (id, expense) => Number(id) === expense.id,
+      from: (to, expense) => new Date(expense.spentAt) > new Date(to),
+      to: (to, expense) => new Date(expense.spentAt) < new Date(to),
+      categories: (category, expense) => category === expense.category,
+    };
+
+    return expenses.filter((expense) =>
+      Object.entries(filters).reduce((result, [currKey, currVal]) => {
```
pedro-ruas, line 35: Even better than using a .reduce here would be using an _every_ instead.
Why? _.reduce_ will iterate and test all query keys always, while an _.every_ will stop iterating once the test fails.
This could improve performance by reducing the amount of iterations/tests

src/models/expenseModel.js
```diff
@@ -0,0 +1,82 @@
...
+    return newExpense;
+  },
+
+  getAllExpenses(filters = {}) {
+    const filteringTests = {
+      userId: (id, expense) => Number(id) === expense.userId,
+      id: (id, expense) => Number(id) === expense.id,
+      from: (to, expense) => new Date(expense.spentAt) > new Date(to),
+      to: (to, expense) => new Date(expense.spentAt) < new Date(to),
+      categories: (category, expense) => category === expense.category,
```
pedro-ruas, line 31: Add a _default_ case here

src/models/expenseModel.js
```diff
@@ -0,0 +1,82 @@
...
+    };
+
+    return expenses.filter((expense) =>
+      Object.entries(filters).reduce((result, [currKey, currVal]) => {
+        const test = filteringTests[currKey];
```
pedro-ruas, line 36: With the _default_ case added, use this:
```
const test = filteringTests[currKey] ?? filteringTests.default;

return test(currVal, expense)
```

src/models/expenseModel.js
```diff
@@ -0,0 +1,82 @@
...
+
+        return result && test
+          ? test(currVal, expense)
+          : currVal === expense[currKey];
+      }, true),
+    );
+  },
+
+  getExpenseById(id) {
```
pedro-ruas, line 45: This function isn't needed, as stated in a previous comment

tests/expense.test.js
```diff
@@ -2,14 +2,18 @@
 
 const supertest = require('supertest');
 const { createServer } = require('../src/createServer');
+const { resetUsers } = require('../src/models/userModel');
+const { resetExpenses } = require('../src/models/expenseModel');
 
-describe('Expense', () => {
+describe.only('Expense', () => {
```
pedro-ruas, line 8: Remove this .only

src/models/userModel.js
```diff
@@ -0,0 +1,52 @@
...
+    users.push(newUser);
+
+    return newUser;
+  },
+
+  getAllUsers() {
+    return users;
+  },
+
+  getUserById(id) {
```
pedro-ruas, line 19: This function isn't needed, as stated in a previous comment

pedro-ruas (changes requested):

src/controllers/expenseController.js
eduvinagre: nice!

src/models/expenseModel.js
eduvinagre: agree

eduvinagre:
@pedro-ruas thanks for all your insights, mate. I'll refactor the code accordingly.

EdPirro (approved):
Great job, @eduvinagre! Keep it up 🚀

PR #627, opened 2024-07-07
solution

Allanpt:


src/controller/expense.controller.js
```diff
@@ -0,0 +1,73 @@
...
+  const newExpense = expenseService.createExpense(req.body);
+
+  if (newExpense === null) {
+    return res.status(400).json({
+      message:
+        "'userId','title','amount' and 'category' content MUST BE filled",
+    });
+  }
+
+  if (newExpense === undefined) {
```
pedro-ruas, line 23: For practical reasons, is this case different than the first one? (newExpense === null)

src/model/user.model.js
```diff
@@ -0,0 +1,58 @@
...
+    this._id = newId;
+  }
+
+  /**
+   * @param {string} newName
+   */
+  set name(newName) {
+    this._name = newName;
+  }
+}
```
pedro-ruas, line 56: You could also have a function for adding expenses to the user

src/service/expenses.service.js
```diff
@@ -0,0 +1,129 @@
...
+    );
+  }
+
+  if (to) {
+    filteredExpenses = filteredExpenses.filter(
+      (expense) => expense.spentAt <= to,
+    );
+  }
+
+  return filteredExpenses;
```
pedro-ruas, line 40: All these _if_ approaches work, but could become something hard to maintain and scale in case of more and more properties/queries needed.
This will keep all scalability in a single place and remove multiple _if_ checks. (because your code checks all ifs, even when already passing the first one).

src/service/expenses.service.js
```diff
@@ -0,0 +1,129 @@
...
+};
+
+const createExpense = (expense) => {
+  const userId = expense.userId;
+  const title = expense.title;
+  const amount = expense.amount;
+  const category = expense.category;
+  const note = expense.note;
+  const spentAt = expense.spentAt;
```
pedro-ruas, line 49: You could change all these const declarations into a desconstucturing:
```
const {
    userId,
    title,
    amount,
    category,
    note,
    spentAt,
} = expense;
```

src/service/expenses.service.js
```diff
@@ -0,0 +1,129 @@
...
+
+  const findUser = getUserById(userId);
+
+  if (!findUser) {
+    return undefined;
+  }
+
+  if (expenses.length !== 0) {
+    newExpense = {
+      id: expenses[expenses.length - 1].id + 1,
```
pedro-ruas, line 65: Seems like you can just use:
`expenses.length + 1`

src/service/expenses.service.js
```diff
@@ -0,0 +1,129 @@
...
+      note,
+    };
+  }
+
+  expenses.push(newExpense);
+
+  return newExpense;
+};
+
+const getExpenseById = (id) => {
```
pedro-ruas, line 92: This function can be replaced by _getExpenses_ if you use the right query

src/service/user.service.js
pedro-ruas, line 1: Basically all comments from expense.service can be used here too xD

pedro-ruas (changes requested):

src/service/expenses.service.js
Allanpt: i couldn't do this lolool. Let's imagine that a have id's 1, 2, 3, 4, 5. If I delete id '3' the array length will be four(4) and the entities that stayed are 1, 2, 4, 5. Then, if I create another entity, the id number will be five(5) because array length is 4, but there is already exists an entity with id 5 you know?

I thought about it and I chose to get the last entity id and add plus one based on it.

`array[array.length - 1].id + 1`

src/controller/expense.controller.js
```diff
@@ -0,0 +1,73 @@
...
+    return res.status(400).json({
+      message: 'Bad Request, user not found',
+    });
```
jv-aquino, line 26: i think, as Pedro, that this is unnecessary

jv-aquino (approved):
Gj Allan!

src/service/expenses.service.js
```diff
@@ -3,50 +3,36 @@ const { Expenses } = require('../model/expenses.model');
...
+  };
+
+  return expensesAsArr.filter((expense) => {
+    return queries.every(([queryKey, queryVal]) => {
+      const filteredFunc =
+        filteringFunctions[queryKey] ?? filteringFunctions.default;
+
+      return filteredFunc(queryVal, expense, queryKey);
+    });
+  });
```
jv-aquino, line 31: nice!

PR #628, opened 2024-07-08
solution

denyssheremeta:


src/services/user.services.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+let users = [];
+
+const getAll = () => users;
+
+const create = (name) => {
+  const newUser = {
+    id: users.length ? users[users.length - 1].id + 1 : 1,
```
DanilWeda, line 9: Create function like a generateId or something like this

src/services/expense.services.js
```diff
@@ -0,0 +1,89 @@
...
+};
+
+const getById = (id) => {
+  const idNum = Number(id);
+
+  return expenses.find((exp) => exp.id === idNum) ?? null;
+};
+
+const create = (data) => {
+  const newId = expenses.length ? expenses[expenses.length - 1].id + 1 : 1;
```
DanilWeda, line 52: Use or create function like a generateId or something like this

src/services/expense.services.js
```diff
@@ -0,0 +1,89 @@
+'use strict';
+
+let expenses = [];
```
DanilWeda, line 3: You can use const in this case, then operate with array methods

src/services/user.services.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+let users = [];
```
DanilWeda, line 3: You can use const in this case, then operate with array methods

src/routes/user.router.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const express = require('express');
+const userController = require('../controllers/user.controllers');
+
+const router = express.Router();
+
+router
+  .get('/', userController.getAllUsers)
```
DanilWeda, line 9: You can move all routes in constants

src/routes/expense.router.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+const express = require('express');
+const expenseController = require('../controllers/expense.controllers');
+
+const router = express.Router();
+
+router
+  .get('/', expenseController.getAllExpenses)
```
DanilWeda, line 9: You can move all routes to constants

src/controllers/user.controllers.js
```diff
@@ -0,0 +1,72 @@
...
+  const { name } = req.body;
+
+  const user = userService.getById(id);
+
+  if (!user) {
+    return res.status(404).send({ error: 'User not found.' });
+  }
+
+  if (!name) {
+    return res.status(400).send({ error: 'Name is required.' });
```
DanilWeda, line 58: Move all status codes and messages to constants

DanilWeda (changes requested):
Good job. 
Check some small comments. 
Many thanks)

DanilWeda (approved):
Good job. 
Approved! 
Many thanks for fixes)

PR #630, opened 2024-07-08
Develop

mfcastilho:


src/repositories/expenseRepository.js
```diff
@@ -0,0 +1,95 @@
...
+        : categories.split(',').map((cat) => cat.trim().toLowerCase());
+
+      if (normalizedCategories.length > 0) {
+        expenses = expenses.filter((expense) =>
+          normalizedCategories.includes(expense.category.toLowerCase()),
+        );
+      }
+      /* eslint-enable */
+    }
+
```
pedro-ruas, line 55: All these if approaches work, but could become something hard to maintain and scale in case of more and more properties/queries needed.
This will keep all scalability in a single place and remove multiple if checks. (because your code checks all ifs, even when already passing the first one).

src/repositories/expenseRepository.js
```diff
@@ -0,0 +1,95 @@
...
+    return expenses;
+  },
+
+  findByPk: (expenseId) => {
```
pedro-ruas, line 59: What is Pk ?

src/repositories/userRepository.js
```diff
@@ -0,0 +1,50 @@
...
+      ...userData,
+    };
+
+    mockUsers.push(newUser);
+
+    return newUser;
+  },
+
+  findAll: () => {
+    const users = mockUsers;
```
pedro-ruas, line 16: Necessary?

src/repositories/userRepository.js
```diff
@@ -0,0 +1,50 @@
...
+
+    return users;
+  },
+
+  findByPk: (userId) => {
```
pedro-ruas, line 21: What is Pk?

src/routes/userRoutes.js
```diff
@@ -0,0 +1,42 @@
...
+  verifyIfUserExists,
+  GetUserController.handle,
+);
+userRoutes.post('/', verifyUserFields, CreateUserController.handle);
+
+userRoutes.patch(
+  '/:id',
+  verifyIfIdIsANumber,
+  verifyIfUserExists,
+  verifyUserFields,
```
pedro-ruas, line 31: _verifyUserFields_ should come before _verifyIfUserExists_.
Why? Because if this verification fails, there is no need to access and search the DB (which is done by _verifyifUserExists_)

src/services/expense/CreateExpenseService.js
pedro-ruas, line 1: There is no problem with your approach, but I don't think there is actually a need to separate every function in a new service file.
There could be just ExpenseService and UserService.
Just a comment, you don't have to change it, don't worry.

But, if all services need to have an _execute_ method, it is a good approach to create a _ServiceInterface_ declaring it and making services extend it

tests/expense.test.js
```diff
@@ -2,14 +2,18 @@
 
 const supertest = require('supertest');
 const { createServer } = require('../src/createServer');
+const userRepository = require('../src/repositories/userRepository');
+const expenseRepository = require('../src/repositories/expenseRepository');
 
-describe('Expense', () => {
+describe.only('Expense', () => {
```
pedro-ruas, line 8: Remove .only

src/routes/userRoutes.js
```diff
@@ -0,0 +1,42 @@
...
+  '/:id',
+  verifyIfIdIsANumber,
+  verifyUserFields,
+  verifyIfUserExists,
+  EditUserController.handle,
+);
+
+userRoutes.delete(
+  '/:id',
+  verifyIfIdIsANumber,
```
jv-aquino, line 37: same comment as pedro did here

PR #637, opened 2024-07-09
solution

alexrapin:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,98 @@
...
+
+const getAllExpenses = (req, res) => {
+  const expenses = expenseService.getAllExpenses(req.query);
+
+  res.statusCode = 200;
+  res.send(expenses);
+};
+
+const getExpenseForPerson = (req, res) => {
+  const { id } = req.params;
```
anastasiiavorobiova, line 12: Consider creating a middleware folder and placing middleware here for checking and validating the user input from the request before passing it to the controller. The user data from the request should be sanitized. Also, you will be able to send 400 errors with additional information about the.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,98 @@
...
+  const expense = expenseService.getExpenseById(id);
+
+  if (!expense) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  res.statusCode = 200;
+  res.send(expense);
```
anastasiiavorobiova, line 22:
```suggestion
try {
  const expense = expenseService.getExpenseById(id);

  if (!expense) {
    res.sendStatus(404);

    return;
  }

  res.statusCode = 200;
  res.send(expense);
} catch (err) {
    res.statusCode = 500;
    res.send('Something went wrong);
}
```
Also, it would be better to wrap this bunch of code inside the try-catch block to handle errors related to the DB connection, etc. In this case, we should provide a client with a proper error code and avoid throwing an error object directly to the client.

anastasiiavorobiova: Also, try to implement these suggestions for all of your code

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,98 @@
...
+  //   typeof userId !== 'number' ||
+  //   typeof title !== 'string' ||
+  //   typeof amount !== 'number' ||
+  //   typeof category !== 'string' ||
+  //   typeof note !== 'string'
+  // ) {
+  //   res.sendStatus(400);
+
+  //   return;
+  // }
```
anastasiiavorobiova, line 61:
```suggestion

```

src/routes/users.route.js
```diff
@@ -0,0 +1,16 @@
+const express = require('express');
+const userController = require('../controllers/users.controller');
+
+const usersRouter = express.Router();
+
+usersRouter.get('/', userController.getAllUsers);
```
anastasiiavorobiova, line 6: Middlewares should be placed here 
```suggestion
usersRouter.get('/', yourMiddlewareToCheckAndSanitizeTheRequest, userController.getAllUsers);
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

lerastarynets (approved):
lgtm 🚀

PR #638, opened 2024-07-09
sol

yepolotn1ak:


src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,62 @@
+const { expensesService } = require('../services/expenses.service');
+const { usersService } = require('../services/users.service');
+
+class ExpensesController {
+  getAllExpenses = (req, res) => {
+    res.send(expensesService.getAll(req.query));
+  };
+  createExpense = (req, res) => {
+    const { userId } = req.body;
```
anastasiiavorobiova, line 9: Consider creating a middleware folder and placing middleware here for checking and validating the user input from the request before passing it to the controller. The user data from the request should be sanitized. Also, you will be able to send 400 errors with additional information about the.

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,62 @@
...
+    const user = usersService.getById(userId);
+
+    if (!user) {
+      res.sendStatus(400);
+
+      return;
+    }
+
+    return res.status(201).json(expensesService.create(req.body));
```
anastasiiavorobiova, line 18:
```suggestion
try {
    const user = usersService.getById(userId);

    if (!user) {
      res.sendStatus(404);

      return;
    }

    return res.status(201).json(expensesService.create(req.body));
} catch (err) {
    res.statusCode = 500;
    res.send('Something went wrong);
}
```
Also, it would be better to wrap this bunch of code inside the try-catch block to handle errors related to the DB connection, etc. In this case, we should provide a client with a proper error code and avoid throwing an error object directly to the client.

anastasiiavorobiova: Also, try to implement these suggestions for all of your code

src/routes/expenses.route.js
```diff
@@ -0,0 +1,12 @@
+const express = require('express');
+const { expensesController } = require('../controllers/expenses.controller');
+
+const expensesRouter = express.Router();
+
+expensesRouter.get('/', expensesController.getAllExpenses);
+expensesRouter.post('/', expensesController.createExpense);
```
anastasiiavorobiova, line 7:
```suggestion
expensesRouter.post('/', yourMiddlewareToCheckAndSanitizeTheRequest, expensesController.createExpense);
```
Middlewares should be placed here

anastasiiavorobiova (changes requested):
Great work! Almost done!

lerastarynets (approved):
lgtm 🚀

PR #640, opened 2024-07-11
Task: Accounting App

HigorPedroso:


src/services/expense.service.js
```diff
@@ -0,0 +1,79 @@
+function getRandomNumber() {
+  const min = 0;
+  const max = 100;
+
+  return Math.floor(Math.random() * (max - min + 1)) + min;
+}
+
+let expenses = [];
+
+const start = () => {
```
jv-aquino, line 10: I think this would be a reset() funcion

src/services/user.service.js
```diff
@@ -0,0 +1,49 @@
+let users = [];
+
+function getRandomNumber() {
+  const min = 0;
+  const max = 100;
+
+  return Math.floor(Math.random() * (max - min + 1)) + min;
+}
```
jv-aquino, line 8: you are reusing this function, don't you think there's a better way to optimize this?

jv-aquino (approved):
Great job Higor!

PR #642, opened 2024-07-11
Solution

MarchAlbion:


src/services/user.service.js
```diff
@@ -0,0 +1,45 @@
+let users = [];
```
DanilWeda, line 1: You can use const, then make operation with array

src/services/user.service.js
```diff
@@ -0,0 +1,45 @@
...
+}
+
+const getAllUsers = () => users;
+
+const getUserById = (id) =>
+  users.find((user) => Number(user.id) === Number(id));
+
+const addUser = (name) => {
+  const newUser = {
+    id: Math.round(Math.random() * 200),
```
DanilWeda, line 14: You can create generateId func for it, and use it in all places in your app

src/services/expenses.service.js
```diff
@@ -0,0 +1,99 @@
+let expenses = [];
```
DanilWeda, line 1: Also you can use const in this case

src/routes/userRouter.js
```diff
@@ -0,0 +1,16 @@
+const express = require('express');
+const userRouter = express.Router();
+
+const userController = require('../controllers/user.controller');
+
+userRouter.get('/', userController.getUsersController);
```
DanilWeda, line 6: You can create constants for all routes

src/routes/expenseRouter.js
```diff
@@ -0,0 +1,19 @@
+const express = require('express');
+const expenseRouter = express.Router();
+const expenseController = require('../controllers/expenses.controller');
+
+expenseRouter.get('/', expenseController.getExpensesController);
```
DanilWeda, line 5: Also, you can create constants for all routes

src/controllers/user.controller.js
```diff
@@ -0,0 +1,77 @@
...
+
+  if (!user) {
+    res.sendStatus(404);
+
+    return;
+  }
+
+  userService.deleteUser(id);
+
+  res.sendStatus(204);
```
DanilWeda, line 68: Move all status code to constants

src/controllers/user.controller.js
```diff
@@ -0,0 +1,77 @@
...
+  if (id === undefined) {
+    res.status(400).send('id is required');
+
+    return;
+  }
+
+  const user = userService.getUserById(id);
+
+  if (user === undefined) {
+    res.status(404).send('Not Found');
```
DanilWeda, line 21: Also move all messages to constants

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,105 @@
+const expenseService = require('./../services/expenses.service');
+const userService = require('./../services/user.service');
+
+function getExpensesController(req, res) {
+  const { userId, categories, from, to } = req.query;
+
+  if (Object.keys(req.query).length !== 0) {
```
DanilWeda, line 7:
```suggestion
  if (Object.keys(req.query).length) {
```
as alternative

DanilWeda (changes requested):
Good job. 
Check some small comments for improving your solution. 
Many thanks)

PR #644, opened 2024-07-16
Solution

dyakym:


src/services/users.service.js
```diff
@@ -0,0 +1,56 @@
+let users = [];
+
+const resetUsers = () => {
+  users = [];
+};
+
+const findUserById = (id) => users.find((user) => user.id === id);
+
+const generateUserId = () =>
+  users.length ? Math.max(...users.map((user) => user.id)) + 1 : 1;
```
DanilWeda, line 10: You can move it in separate file and use in all places where you need id

PR #647, opened 2024-07-22
add task solution

lmuias:
Just a question of curiosity, is it possible to do this task with using uuid library for id generation, because i tried to do such thing, but got problem with matching id`s( i guess its because uuid in complicated format).


SanyaBratashchuk (changes requested):
In general you can use uuidV4, but tests expects that id of entities is number but uuid returns string

SanyaBratashchuk (approved):

PR #654, opened 2024-08-08
Task Completed

Yaroslav-Kononenko:


src/createServer.js
```diff
@@ -1,11 +1,24 @@
...
-  // Return the server (express app)
+  const app = express();
+
+  app.use(cors());
+
+  resetUsers();
+  resetExpenses();
+
+  app.use('/users', express.json(), usersRouter);
+  app.use('/expenses', express.json(), expensesRouter);
```
BudnikOleksii, line 19:
```suggestion
  app.use(cors());
  app.use(express.json());

  resetUsers();
  resetExpenses();

  app.use('/users', usersRouter);
  app.use('/expenses', expensesRouter);
```
You can apply middleware to all routes at once

src/services/expenses.services.js
```diff
@@ -0,0 +1,123 @@
...
+ *@typedef {Object} Expense
+ *@property {number} id
+ *@property {number} userId
+ *@property {string} spentAt
+ *@property {string} title
+ *@property {number} amount
+ *@property {string} category
+ *@property {string} [note]
+ *@type {Expense[] | []}
+ */
```
BudnikOleksii, line 15: 👍

src/services/expenses.services.js
```diff
@@ -0,0 +1,123 @@
...
+
+  if (categories) {
+    const categoryList = categories.split(',');
+
+    filteredExpenses = filteredExpenses.filter((expense) => {
+      return categoryList.includes(expense.category);
+    });
+  }
+
+  return filteredExpenses;
```
BudnikOleksii, line 50: Try to filter just once

src/services/expenses.services.js
```diff
@@ -0,0 +1,123 @@
...
+    if (category) {
+      searchedExpense.category = category;
+    }
+
+    if (note) {
+      searchedExpense.note = note;
+    }
+  }
+
+  return searchedExpense;
```
BudnikOleksii, line 94: You can use object.assign instead

src/services/expenses.services.js
```diff
@@ -0,0 +1,123 @@
...
+};
+
+const getExpenseById = (id) => {
+  return findExpenseById(id) || null;
+};
+
+const removeExpenseById = (id) => {
+  const clearedExpenses = expenses.filter((expense) => expense.id !== id);
+
+  const isRemoved = clearedExpenses.length !== expenses.length || null;
```
BudnikOleksii, line 104: Why do you set `null`?

src/services/users.services.js
```diff
@@ -0,0 +1,56 @@
...
+/**
+ *@typedef {Object} User
+ *@property {number} id
+ *@property {string} name
+ *@type {User[] | []}
+ */
+let users = [];
+
+const getAll = () => {
+  return users.length !== 0 ? users : [];
```
BudnikOleksii, line 14:
```suggestion
  return users;
```
It doesn't make much sense

src/services/users.services.js
```diff
@@ -0,0 +1,56 @@
...
+    name,
+  };
+
+  users.push(newUser);
+
+  return newUser;
+};
+
+const updateById = (id, name) => {
+  const searchedUser = users.find((user) => user.id === id);
```
BudnikOleksii, line 43: You have function `getUserById`, use it

src/services/users.services.js
```diff
@@ -0,0 +1,56 @@
...
+
+  const result = searchedUser ? Object.assign(searchedUser, { name }) : null;
```
BudnikOleksii, line 45: If you don't find user you can throw an error and send to the client that user not found

src/services/expenses.services.js
Yaroslav-Kononenko: I wanted to make it more explicit than the usual “undefined”. But I guess that doesn't make sense here

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,65 @@
...
+};
+
+const createExpense = (req, res) => {
+  const { userId, spentAt, title, amount, category, note } = req.body;
+
+  if (!userId || !spentAt || !title || !amount || !category) {
+    return res.status(400).send({ message: 'Missing required fields' });
+  }
+
+  // Check is user exist, if no, throw Error
```
BudnikOleksii, line 17: Remove comments, they are redundant

src/middlewares/middleware.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const errorHandler = (err, req, res, next) => {
+  let statusCode = 500;
+
+  if (err.message === 'Not found') {
+    statusCode = 404;
+  } else if (err.message === 'Bad request') {
+    statusCode = 400;
+  }
```
BudnikOleksii, line 10: To set type depends on the message is not flexible, and in you can you have magic values. 
A simple example of an error handler https://github.com/BudnikOleksii/rick-n-morty-legacy-/blob/main/server/src/middlewares/error-handler.js
https://github.com/BudnikOleksii/rick-n-morty-legacy-/blob/main/server/src/utils/errors/api-errors.js

BudnikOleksii (approved):
Good job 👍

src/middlewares/middleware.js
Yaroslav-Kononenko: Thank you very much, that's helpful)

PR #665, opened 2024-09-22
Add task solution

Eater228:


src/controllers/expenseControlle.js
```diff
@@ -0,0 +1,113 @@
+/* eslint-disable no-useless-return */
+const expenseService = require('../services/expenseService');
+const userService = require('../services/userService');
+
+const getExpenses = (req, res) => {
+  const { userId, categories, from, to } = req.query;
```
anastasiiavorobiova: Consider wrapping your code into a try-catch block and sending a 500 error if something goes wrong on the server

anastasiiavorobiova (changes requested):
Great work! Almost done

SanyaBratashchuk (approved):

PR #675, opened 2024-10-25
task solution

arthurfujii:


src/controllers/user.controller.js
```diff
@@ -0,0 +1,75 @@
+const {
+  getAllUsers,
+  getUserById,
+  createUser,
+  removeUser,
+  updateUser,
+} = require('../services/user.service');
+
+// const userController = {
```
pedro-ruas: Remove comment

pedro-ruas (approved):
I know you will make some changes still, but this is already approved =D 
Noice!

PR #764, opened 2025-05-05
Solution for accounting-app

kovaltar:
It seems that in this task it's impossible to pass all the tests. It looks like this is happening because the tests populate the users and expenses arrays with data, but they don't automatically clear them. As a result, when new tests are run and expect certain data to be missing, that data is actually still present, which leads to different status codes being returned than what the tests expect.


anastasiiavorobiova (approved):
Great work!

PR #767, opened 2025-05-15
soluçao

LuisCampelo2:


pedro-ruas (changes requested):
Nothing actually wrong with your code, but I really think you should improve the structure/organization of it.
It is a terrible idea to keep all code in a single gigantic file. You should split it into smaller files that make sense, for example:
- 1 file for the account controller
- 1 file for the user controller
- 1 file for the user router
- 1 file for the account router
- 1 file for the server itself
... any other needed file

This is a simple project, but in a real project you would easily get lost by doing all in a single file. better improve now than later ;D

PR #796, opened 2025-07-19
Solution

Vasya564:


Alkaponees (changes requested):
It seems to me that is not a task for DevOps course

PR #814, opened 2025-09-09
Add task solution

FlavioMiranDev:


pedro-ruas (changes requested):
Ok, it's well done and works, but a code like this in the real world would be terrible to maintain and work on.
What I mean is that, you should divide this file in a way that makes sense, not just create a single huge file, it is too much code in the same place.

The usual approach would be to create at least a `controllers` folder where you create a class to handle each route.

pedro-ruas (approved):
Nice! This looks much more like a professional code you will use in the real world. Great job!

PR #838, opened 2025-10-17
Implement accounting app with users and expenses endpoints

djoseph-png:
- Add 5 user endpoints (GET, POST, PATCH, DELETE)
- Add 5 expense endpoints (GET, POST, PATCH, DELETE)
- Implement filters for expenses (userId, date range, categories)
- Add validation for required fields (400 error)
- Add validation for non-existent entities (404 error)

### DIFF
--- a/.github/workflows/test.yml-template
+++ b/.github/workflows/test.yml-template
@@ -1,0 +1,23 @@
+name: Test
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [20.x]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm install
+    - run: npm test
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
       },
       "devDependencies": {
         "@mate-academy/eslint-config": "latest",
-        "@mate-academy/scripts": "^1.8.6",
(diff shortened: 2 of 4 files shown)


djoseph-png:
@mateacademy-ai-mentor 
## Response to Mentor Feedback: Status Code Clarification

### Context

During the review process, there was a discrepancy between the mentor's feedback and the automated test expectations regarding the appropriate HTTP status code when a referenced user is not found during expense creation (`POST /expenses`).

### The Issue

**Mentor's Request:**
- Return `404 Not Found` when the referenced user does not exist
- Rationale: Missing entities should return 404 per REST conventions

**Automated Test Expectation:**
- The test `should return 400 if user not found` explicitly expects `400 Bad Request`
- The test fails when returning 404: `expected 400 "Bad Request", got 404 "Not Found"`

### Analysis

After careful consideration, I've implemented the solution that satisfies the **automated test requirements** (returning `400`) for the following reasons:

1. **Test-Driven Development:** The automated tests define the contract and expected behavior for this application
2. **Validation Context:** In this specific case, providing a non-existent `userId` could be interpreted as a validation error (invalid reference), which justifies a `400` response
3. **Consistency:** The implementation passes all 24 automated tests, ensuring the application meets all specified requirements

### Implementation Details

The current implementation in `POST /expenses`:

```javascript
// Validate required fields first
if (userId === undefined || userId === null || ...) {
  return res.status(400).json({ error: 'Missing required fields' });
}

// Normalize userId to number
const userIdNum = Number(userId);

// Validate userId is a valid number
if (isNaN(userIdNum)) {
  return res.status(400).json({ error: 'Invalid userId' });
}

// Check if user exists - return 400 per test expectation
const user = users.find((u) => u.id === userIdNum);
if (!user) {
  return res.status(400).json({ error: 'User not found' });
}
```
image

PR #840, opened 2025-10-22
added solution

dopezinho:


pedro-ruas (changes requested):
So, this code works, but it is too much code in a single file.
In a real project, this is not just bad practice, but wouldn't be approved if working on a team with other devs.
Please create a structure to divide this code in a way that makes sense. We usually follow:
- main/server file (entry point and server declaration
- routes: one route for each path (expenses, users)
- controllers: one for each path (userController, expensesController)

pedro-ruas (approved):
Now we are talking!

PR #843, opened 2025-11-07
Develop

isapq:


pedro-ruas (changes requested):
Please, separate this project in more files.
In a real world project, you would not keep all the code in "server.js", so it's better to exercise that.
Create a separation that makes sense, with a folder for user and other for expenses

pedro-ruas (approved):

PR #853, opened 2025-12-26
Develop

mForkosh:


src/services/users.service.js
```diff
@@ -0,0 +1,69 @@
...
+
+      break;
+    }
+  }
+
+  const newUser = {
+    id: accessibleId,
+    name,
+  };
+
```
Anton-Kuchmasov, line 34: just use `users.length + 1` - for this task it'll be enough

in prod projects, use only auto-generated IDs e.g. [uuid](https://www.npmjs.com/package/uuid)

src/services/users.service.js
```diff
@@ -0,0 +1,69 @@
...
+};
+
+const deleteById = (id) => {
+  const index = users.findIndex((u) => u.id === id);
+
+  if (index === -1) {
+    return null;
+  }
+
+  return users.splice(index, 1);
```
Anton-Kuchmasov, line 47: `splice` is mutable method, so you don't need to return whole mutated array

in CRUD standards, DELETE endpoint can return 204 or deleted (full/partial) entity

Anton-Kuchmasov (approved):
Good job!

Check out comments below

PR #856, opened 2026-01-04
solution

Ga1dar:


brespect (changes requested):
Good progress, but let`s split the App into different logical components

PR #858, opened 2026-01-06
Solution node_accounting-app-v2

RafaelGratton:


IanLuan (approved):
Nice job! One tip: next time, try to split the logic into more files to make it easier to maintain.

PR #860, opened 2026-01-08
solution

nadiaaamf:
# Accounting app (with Node.js)


IanLuan (approved):
Good job! One tip: split your logic into more files to make it easier to maintain. Keeping everything in a single file is not a good practice.

PR #864, opened 2026-01-14
node_accounting-app-v2 Develop

DemetriodosAnjos:
Implementação completa de Users e Expenses + testes passando (24/24)


pedro-ruas (changes requested):
Ok, looks good, but the code should be split into more files.
Ina real-world project, you would never have a `createServer` file this big and handling all routes and controllers logic.
Split it at least into 2 other files: `userController` and `expenseController` with their respective logic

PR #865, opened 2026-01-14
Solution

hma-3:


hma-3:
@mateacademy-ai-mentor I did the task according to tests and [documentation](https://app.swaggerhub.com/apis/YURIIHOLIUK_1/Accounting-App/1.0.0). Both of them describe collections `users` and `expenses`. And as you can see the tests are passing, so I wrote the code for correct collections, right fields for them and right methods (`PATCH`, not `PUT`). If I rewrite the code as you think I should, the tests won't pass, and I won't be able to push my code.

hma-3:
@mateacademy-ai-mentor Additionally, according to the tests and documentation, if a user does not exist when creating a new expense, the API should return a 400 Bad Request (meaning that wrong data was passed).

src/controllers/expenses.controller.js
```diff
@@ -0,0 +1,113 @@
...
+    res.status(404).json({ error: 'Expense not found' });
+
+    return;
+  }
+
+  if (!spentAt && !title && !amount && !category && !note) {
+    res.status(400).json({ error: 'At least one field is required' });
+
+    return;
+  }
```
hma-3, line 93: @mateacademy-ai-mentor I've added validation for fields, but only when there is none of them, since the tests and documentation expect `PATCH` method, not `PUT`

PR #867, opened 2026-01-23
Dev

DrDakka:


src/main.js
```diff
@@ -1,8 +1,9 @@
 'use strict';
 
 const { createServer } = require('./createServer');
+const { PORT } = require('./static/constants');
```
Anton-Kuchmasov, line 4: best practice is using `env` variable for PORT number (constant can be used as a fallback)

src/controllers/controllers.js
```diff
@@ -0,0 +1,21 @@
+const { dbActions, mthds } = require('../static/constants');
+
+const methodsToDbActions = {
+  [mthds.post]: dbActions.post,
+  [mthds.patch]: dbActions.patch,
+  [mthds.delete]: dbActions.delete,
+};
```
Anton-Kuchmasov, line 7: Probably, should be helpful in Typescript :)

src/helpers/helpers.js
```diff
@@ -0,0 +1,62 @@
+// since the only body we can receive, I hardcode JSON parse to the function
+
+async function readBody(req, maxSize = 1048576) {
```
Anton-Kuchmasov, line 3: magic number

also, pay attention for AI comments - Express validates requests for you by default :)

Anton-Kuchmasov (approved):

src/main.js
DrDakka: My bad, thank you for pointing that

src/controllers/controllers.js
DrDakka: It was inspired by it :D

src/helpers/helpers.js
DrDakka: Thanks for pointing it out as well :)

The whole thing was written w/o express, so validation was needed :D

PR #872, opened 2026-02-02
 node_accounting-app-v2_TetSh

TetiankaSh:


src/services/users.services.js
```diff
@@ -0,0 +1,15 @@
+const findUser = (users, id) => {
+  return users.find((u) => u.id === +id);
+};
+
+const filterUsers = (users, id) => {
+  // eslint-disable-next-line no-param-reassign
+  users = users.filter((u) => u.id !== +id);
```
FaiHamid: Here you should create separate variable where put filter result.
Remove comment!
Don`t use one letter name for variables!
It should look like this:
`const result = users.filter(user => user.id != +id)
return result;
`

src/services/users.services.js
```diff
@@ -0,0 +1,15 @@
+const findUser = (users, id) => {
+  return users.find((u) => u.id === +id);
```
FaiHamid: Same here don`t use one letter names.

src/services/expenses.services.js
```diff
@@ -0,0 +1,15 @@
+const findExpense = (expenses, id) => {
+  return expenses.find((exp) => exp.id === +id);
+};
+
+const filterExpense = (expenses, id) => {
+  // eslint-disable-next-line no-param-reassign
+  expenses = expenses.filter((exp) => exp.id !== +id);
```
FaiHamid: Look at changes that I wrote for users.service!

src/controllers/users.controller.js
```diff
@@ -0,0 +1,97 @@
...
+
+    return;
+  }
+
+  res.status(200).send(user);
+};
+
+const deleteUser = (users) => (req, res) => {
+  const { id } = req.params;
+  const index = users.findIndex((u) => u.id === +id);
```
FaiHamid: Same here don`t use one letter names.

FaiHamid (changes requested):

FaiHamid (approved):
Good job!

PR #874, opened 2026-02-05
add task solution

nikita4880:


brespect (changes requested):
Good job, but let`s separate your up accourding to MVC model, to routres, services and controllers

brespect (approved):
Perfect 🚀

PR #889, opened 2026-03-08
Implement users and expenses API

marcellealvb:


IanLuan (changes requested):
There are some errors blocking the tests from running:

1. 'users' is already declared and 'filteredUsers' is not defined
The problem here is that you already declared `const users = []` on line 14. Because of that, every time you try to declare const users again or reassign it, this error will occur.

2. 'filteredUsers' is not defined
You are assigning a value to filteredUsers without declaring it first.

Please review these errors and run npm run test to check if everything is working correctly.

pedro-ruas (approved):

PR #895, opened 2026-03-22
feat: implement users and expenses CRUD endpoints

glauccoeng-prog:


src/createServer.js
```diff
@@ -1,11 +1,186 @@
...
+
+  app.use(express.json());
+
+  // Armazenamento em memória (limpo a cada chamada do createServer)
+  const users = [];
+  const expenses = [];
+  let userIdCounter = 0;
+  let expenseIdCounter = 0;
+
+  // ==================== USUÁRIOS ====================
```
raulriato: Instead of dividing the code into sections, your should separate your code in multiple files. Make at least the separation into controller and service, but if you wanna go the best way, create folders for controllers and services and separate them in usersController and expensesController files, as well as for services.

raulriato (changes requested):
Your code looks fine, you just need to divide it better.

src/createServer.js
glauccoeng-prog: Great suggestion! I've refactored the entire file following the controller/service pattern. I also took the liberty to add a routers/ folder to map the endpoints to keep createServer.js as slim as possible.

DanielCamposDevX (approved):

PR #902, opened 2026-03-25
Develop

wzfxc:


brespect (changes requested):
Good progress, let's separate your App using MVC pattern (services, routes, controllers, models). Also, fix all tests before requesting the review

brespect (approved):
Well done!

PR #921, opened 2026-04-17
solution

johnnyvasconcelos:


raulriato (changes requested):
Your code works, but it's not a good practice to have all your code in just one file. Divide your code, create at least a `Controller` and a `Service` for `expenses` and `users`

PR #928, opened 2026-05-01
solution

mykytalandar:


mykytalandar:
User › deleteUser › should return 404 if user does not exist
User › getUser › should return 404 if user does not exist
User › getUsers › should return empty array if no users
Expense › createExpense › should return 400 if user not found

The tests don't pass, although when checked through Postman everything works correctly.

mykytalandar:
new solution


I rewrote the code without using the MVC approach, and all tests pass.

2pasha (approved):
lgtm 🔥

PR #952, opened 2026-06-30
node_accounting-app-v2

Elber-Kevenny:
node_accounting-app-v2


raulriato (changes requested):
Your code seems okay, but it's not a good practice to have all the conde on one single file. I'll ask you to divide your code at least in separated controllers and services, for expenses and users. Ideally, separate the routes too.

raulriato (approved):
Good job! Just a suggestion for your future Node.js projects: as applications become larger, organizing the project into folders with clear responsibilities can make the code easier to navigate and maintain.

A structure like this is a good starting point:

```text
src/
├── controllers/
│   ├── expensesController.js
│   └── usersController.js
├── routes/
│   ├── expensesRoutes.js
│   └── usersRoutes.js
├── services/
│   ├── expensesService.js
│   └── usersService.js
└── createServer.js
```

With this approach:

* **routes** define the API endpoints.
* **controllers** handle the HTTP requests and responses.
* **services** contain the business logic.

Another small recommendation is to keep the file naming consistent with the project's existing convention. Since this project already uses `createServer.js` (camelCase), I would keep following that pattern instead of mixing different naming styles.

It's also worth knowing that you'll see different naming conventions across projects. For example, these are all valid and commonly used:

* `usersService.js` (camelCase)
* `users-service.js` (kebab-case)
* `users.service.js` (dot notation)

The specific convention is less important than being consistent throughout the project.

If you Have any interest in learning more about some of these, feel free to join a `Bate papo com o mentor` meeting and asking about it.

raulriato (approved):

PR #961, opened 2026-07-20
Solution

jeronymomatheus:


src/createServer.js
raulriato, line 1: You should not have all your code in a single file, this is not good practice.

Separate your code at least in a controller and a router layer. Ideally, your should have this structure:
```
src/
├── controllers/
│   ├── expensesController.js
│   └── usersController.js
├── services/
│   ├── expensesService.js
│   └── usersService.js
├── routers/
│   ├── expensesRouter.js
│   └── usersRouter.js
└── createServer.js
```

But the minimum will be:

```
src/
├── controllers/
│   ├── expensesController.js
│   └── usersController.js
├── routers/
│   ├── expensesRouter.js
│   └── usersRouter.js
└── createServer.js

raulriato (changes requested):
Your code works, but it's not clean. Follow the instructions and you'll be good to go.

Also, don't leave commented code in the project.

DanielCamposDevX (approved):

PR #962, opened 2026-07-21
solution

Y-Kotvitskiy:


Y-Kotvitskiy:
I already rename file eslintrc.js to  eslintrc.cjs, but lint test does`t work

PR #968, opened 2026-08-01
Solution

AzartiksMan:
Implement expense tracking app: users & expenses, 5 endpoints each. Layered architecture (routes → controllers → services), in-memory storage, filters on GET /expenses, validation (400/404). All tests pass, ESLint clean.


AzartiksMan:
@mateacademy-ai-mentor

I've fixed point 2 — in `expenses.controller.js` the required-parameter validation now runs **before** the user existence check.

Regarding point 1 (required-field validation in the update endpoints):

- In `users.controller.js` the `update` function **already** validates `name` and returns 400 when it is missing.
- For `expenses`, the PATCH endpoint is intentionally a **partial update**. The provided test `updateExpense › should update expense` sends only `{ title: 'Buy a new TV' }` and expects a **200** with the merged result. Requiring all fields in PATCH would make that official test fail, so full required-field validation cannot be added to the expense update without breaking the test suite.

Could you please re-check with this in mind? Thank you!

